PR #3, opened 2019-04-26
docs: godoc for package and example

DenBeke:


example_test.go
```diff
@@ -0,0 +1,32 @@
+package scalewaysdkgo
+
+import (
+	"fmt"
+
+	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+func Example_apiClient() {
```
kindermoumoute: Let's keep things simple: no error handling, no pretty display.

Also we could add few comment on what does each line.

```Go
func Example_apiClient() {
	// create a Scaleway client
 	client, _ := scw.NewClient(
		scw.WithAuth("ACCESS_KEY", "SECRET_KEY"), // Get your credentials at https://console.scaleway.com/account/credentials
	)

	// get instance SDK
 	instanceApi := instance.NewApi(client)

	// do a request
 	response, _ := instanceApi.ListServers(&instance.ListServersRequest{
		Zone: scw.ZoneFrPar1,
	})

	fmt.Println(response)
 }
```

example_test.go
```diff
@@ -0,0 +1,32 @@
...
+	instanceApi := instance.NewApi(client)
+
+	response, err := instanceApi.ListServers(&instance.ListServersRequest{
+		Zone: scw.ZoneFrPar1,
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	for _, server := range response.Servers {
```
kindermoumoute: Is it required to have valid Go code in example? ie. can we not use response here?

DenBeke: The example should compile.

example_test.go
DenBeke: I just copied the file from the `README.md` file.

kindermoumoute: - create client (simple)
- list servers
- create load balancer

example_test.go
```diff
@@ -0,0 +1,83 @@
...
+	// Create SDK objects for specific Scaleway Products
+	instance := instance.NewApi(client)
+	lb := lb.NewApi(client)
+
+	// Start using the SDKs
+	_, _ = instance, lb
+
+}
+
+func Example_getServers() {
```
kindermoumoute: list servers*

PR #4, opened 2019-04-30
feat: add ServiceInfo type

jerome-quere:


utils/service_info.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+// ServiceInfo contains an API information.
+// This information is only here for debugging. Do not really on this value for your project.
```
kindermoumoute: Do not rely*

kindermoumoute: these values

utils/service_info.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+// ServiceInfo contains an API information.
```
kindermoumoute: contains service data?

PR #5, opened 2019-05-02
feat: change visibility of errorResponse struct

Sh4d1:
Changed visibility of this types to be able to cast an error to this type and thus getting the status code.

### DIFF
--- a/scw/errors.go
+++ b/scw/errors.go
@@ -6,7 +6,8 @@ import (
 	"net/http"
 )
 
-type errorResponse struct {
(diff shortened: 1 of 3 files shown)


scw/errors.go
```diff
@@ -6,7 +6,8 @@ import (
 	"net/http"
 )
 
-type errorResponse struct {
+// ErrorResponse is an error type for the Scaleway API
+type ErrorResponse struct {
```
kindermoumoute: let's rename it to `ResponseError`

PR #7, opened 2019-05-03
feat: implement default field ClientOptions

QuentinBrosse:
Implemented three new `ClientOption`:

- `WithDefaultOrganizationId` client option sets the client default organization ID.  
It will be used as the default value of the organization_id field in all requests made with this client.
- `WithDefaultRegion` client option sets the client default region.  
It will be used as the default value of the region field in all requests made with this client.
- `WithDefaultZone` client option sets the client default zone.  
It will be used as the default value of the zone field in all requests made with this client.

### DIFF
--- a/scw/client.go
+++ b/scw/client.go
@@ -68,6 +68,18 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 	}, nil
 }
 
+func (c *Client) GetDefaultOrganizationId() string {
+	return c.defaultOrganizationId
+}
+
+func (c *Client) GetDefaultRegion() Region {
+	return c.defaultRegion
+}
+
+func (c *Client) GetDefaultZone() Zone {
+	return c.defaultZone
+}
+
 func defaultOptions() []ClientOption {
 	return []ClientOption{
 		WithApiUrl("https://api.scaleway.com"),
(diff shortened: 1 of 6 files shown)


scw/client.go
```diff
@@ -15,10 +15,13 @@ import (
 // This client should be passed in the `NewApi` functions whenever an API instance is created.
 // Creating a Client is done with the `NewClient` function.
 type Client struct {
-	httpClient *http.Client
-	auth       auth.Auth
-	baseUrl    string
-	userAgent  string
+	httpClient            *http.Client
+	auth                  auth.Auth
+	baseUrl               string
```
jerome-quere: rename api_url to be iso with future config file

scw/client_test.go
```diff
@@ -38,6 +41,9 @@ func TestNewClientWithOptions(t *testing.T) {
 		WithEndpoint(testEndpoint),
```
jerome-quere: Rename `WithApiUrl`

scw/settings.go
```diff
@@ -9,12 +9,15 @@ import (
 )
 
 type settings struct {
-	Url           string
-	Token         auth.Auth
-	DefaultRegion string
-	UserAgent     string
-	HttpClient    *http.Client
-	Insecure      bool
+	Url        string
```
jerome-quere: Rename to `ApiUrl`

PR #10, opened 2019-05-03
fix: compare nil value with reflect.Value.IsNil

Sh4d1:
Fix bug for when a pointer is not initialized

### DIFF
--- a/internal/parameter/query.go
+++ b/internal/parameter/query.go
@@ -8,10 +8,12 @@ import (
 
 // AddToQuery add a key/value pair to an URL query
 func AddToQuery(query url.Values, key string, value interface{}) {
-	if value == nil {
+	elemValue := reflect.ValueOf(value)
+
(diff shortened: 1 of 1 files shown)


internal/parameter/query.go
```diff
@@ -8,10 +8,12 @@ import (
 
 // AddToQuery add a key/value pair to an URL query
 func AddToQuery(query url.Values, key string, value interface{}) {
-	if value == nil {
+	elemValue := reflect.ValueOf(value)
+
+	if elemValue.Kind() == reflect.Ptr && elemValue.IsNil() {
```
kindermoumoute: Update the condition to check `reflect.Invalid`
```Go
if elemValue.Kind() == reflect.Invalid || elemValue.Kind() == reflect.Ptr && elemValue.IsNil() {
```

Sh4d1: :+1:

PR #15, opened 2019-05-09
feat: add list of regions and zones

Sh4d1:

### DIFF
--- a/utils/locality.go
+++ b/utils/locality.go
@@ -1,18 +1,43 @@
 package utils
 
-// Region is a geographical location
-type Region string
-
-const (
-	RegionFrPar = Region("fr-par")
-	RegionNlAms = Region("nl-ams")
-)
-
 // Zone is an availability zone
 type Zone string
 
 const (
+	// ZoneFrPar1 represents the fr-par-1 zone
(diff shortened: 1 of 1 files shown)


scw/region.go
```diff
@@ -15,3 +15,21 @@ const (
 	ZoneFrPar1 = Zone("fr-par-1")
```
QuentinBrosse: Please add the `fr-par-2` zone.

Sh4d1: done

QuentinBrosse:
I just moved `scw/region.go` to [`utils/locality.go`](https://github.com/scaleway/scaleway-sdk-go/blob/master/utils/locality.go) in #18 to avoid a circular dependancy.

Can you rebase your PR on `master` and refactor it, please?
I can handle it if you prefer. ;)

Thanks!

Sh4d1:
Done :)

QuentinBrosse:
We prefer:
- `AllRegions` instead of `Regions` 
- Replace `ZonesInRegion` constant by this getter: `func (r *Region) GetZones() []*Zone`
Ca irait pour ton use case ?

Sh4d1:
:+1:  juste changed pointers to normal string

PR #23, opened 2019-05-13
Fix: content length

### LINKED ISSUE #24: Can't patch or post on instance
Version: f1fb457

When posting or patching on instancem I got:
`received status code 400: error type is invalid_request_error: Request data is empty. `

kindermoumoute:
Fix #24 

Content length was not set in default request. It was using a custom Reader (internal/marshaler.Body) which is not known by [http.NewRequest](https://golang.org/src/net/http/request.go?s=26724:26793#L834), causing the request to have no ContentLength.


Archie1978:
Hello, I have a same connective. It is very good kindermoumoute.

PR #26, opened 2019-05-14
fix: clean up internal package

kindermoumoute:
* remove marshal body function
* add tests for SetBody method


scw/request_test.go
```diff
@@ -83,3 +85,39 @@ func TestGetHeadersWithBody(t *testing.T) {
 
 	testhelpers.Equals(t, expectedHeaders, allHeaders)
 }
+
+func TestSetBody(t *testing.T) {
```
DenBeke: Wouldn't it make sense to also test `SetBody` with a file? Because afaik this is not yet covered.

scw/request_test.go
```diff
@@ -83,3 +85,66 @@ func TestGetHeadersWithBody(t *testing.T) {
...
+	r, isBytesReader := req.Body.(*bytes.Reader)
+
+	testhelpers.Assert(t, isBytesReader, "req.Body should be bytes Reader")
+
+	b := make([]byte, r.Len())
+	_, err := r.Read(b)
+	testhelpers.Ok(t, err)
+
+	testhelpers.Equals(t, []string{"plain/text"}, req.Headers["Content-Type"])
+	testhelpers.Equals(t, `some body`, string(b))
```
DenBeke: Why don't you use `testBody` body here instead of `some body` string?

PR #28, opened 2019-05-14
feat: log things

kindermoumoute:


scw/client.go
```diff
@@ -55,7 +56,10 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 		if clientTransport.TLSClientConfig == nil {
 			clientTransport.TLSClientConfig = &tls.Config{}
 		}
+		logger.Debugf("use TLS mode")
```
DenBeke: I'd write `using`.

scwconfig/load.go
```diff
@@ -34,13 +36,15 @@ func Load() (Config, error) {
 		}
 		confV1, err := unmarshalConfV1(content)
 		if err == nil {
+			logger.Warningf("loaded config V1 from %s: config V1 is deprecated, please switch your config file to the V2", configPath)
 			return confV1.toV2().catchInvalidProfile()
 		}
 		confV2, err := unmarshalConfV2(content)
```
QuentinBrosse: You don't need to catch the `error` of `unmarshalConfV1` before unmarshal the v2 config? It could lead to a v2 error message for a v1 config.

kindermoumoute: This error happens when file does not exist. At this point in the code we want to try for V2 before returning an error.

scwconfig/load.go
```diff
@@ -34,13 +36,15 @@ func Load() (Config, error) {
 		}
 		confV1, err := unmarshalConfV1(content)
 		if err == nil {
+			logger.Warningf("loaded config V1 from %s: config V1 is deprecated, please switch your config file to the V2", configPath)
```
jerome-quere: If the user see this he will be like OK but What should I do ? Isn't migration automatic ?

kindermoumoute: It would be nice to add a link to a quick doc (e.g.: `https://github.com/scaleway/scaleway-sdk-go/blob/master/scwconfig/README.md`)

scwconfig/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```yaml
+# get your access and secret keys on https://console.scaleway.com/account/credentials
+access_key: SCWXXXXXXXXXXXXXXXXX
+secret_key: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+default_organization_id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+default_region: fr-par
+default_zone: fr-par-1
+```
+
```
QuentinBrosse: Can you add a list of the priority order, for ex: (a low number have a high priority)
1. varenv
2. deprecated var env
3. config file (v2)
4. deprecated config file (v1) If this file is found an automatic migration (v1 => v2) is made.

PR #33, opened 2019-05-15
Create and modifie instance

Archie1978:
Fix the update of the instance and the modification of an ip.
I added an example in the readme.


DenBeke:
Dear @Archie1978,

We are really happy to see you contributing to our new Go SDK, but unfortunately we cannot accept this pull request.
You made changes to an auto-generated file, which means that they will be discarded once we change something in the API definitions at our side (this happens a lot).

meganmcclendon:
Well my 40 dollars was yanked off my card and it brought me to your sight

PR #40, opened 2019-05-17
style: better http error message

jerome-quere:
Improve Http error string message:
```
panic: http request [GET https://api.scaleway.com/instance/v1/zones/fr-par-1/servers?organization=] return [400 Bad Request]
{
  "message": "Validation Error",
  "type": "invalid_request_error",
  "fields": {
    "organization": [
      " is not a valid UUID."
    ]
  }
}
```

### DIFF
--- a/scw/errors.go
+++ b/scw/errors.go
@@ -19,6 +19,9 @@ type ResponseError struct {
 
 	// StatusCode is the HTTP status code received
 	StatusCode int `json:"-"`
+
+	// Status is the HTTP status received
+	Status string `json:"-"`
 }
 
 func hasResponseError(res *http.Response) error {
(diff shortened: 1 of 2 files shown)


scw/errors.go
```diff
@@ -44,18 +57,12 @@ func hasResponseError(res *http.Response) error {
...
 
 func (e *ResponseError) Error() string {
-	s := fmt.Sprintf("received status code %d", e.StatusCode)
+	s := fmt.Sprintf("http request [%s %s] return [%s]", e.RequestMethod, e.RequestUrl, e.Status)
 
-	if e.Type != "" {
-		s = fmt.Sprintf("%s: error type is %s", s, e.Type)
+	jsonStr, _ := json.MarshalIndent(e, "", "  ")
+	if len(jsonStr) > 0 {
+		s = fmt.Sprintf("%s\n%s", s, jsonStr)
```
kindermoumoute: an error should not be multi-line, maybe use a log.Debugf in `hasResponseError` to show print full http body

PR #43, opened 2019-05-17
feat: update generated apis

### LINKED ISSUE #12: Error when listing load balancers
SDK version: master (8cf099e)
Go version: 1.12.4
GOARCH: amd64
GOOS: linux

When I call :

```go
resp, err := l.api.ListLbs(&lb.ListLbsRequest{
  Region:   l.region,
  Name:     &lbName,
  PageSize: &pageSize,
 Page:     &page,
})
```

I get 
```
received status code 400:  is not a valid lb.ListLbsRequest_OrderBy
```

### LINKED ISSUE #25: Override omitempty
Hello,
     If you use the API, you notice that all the arguments are omitempty. So, I can not detach an IP from an instance.
I propose to create an override attribute and to do the json code on a map [string] interface {} for each request.
Example:
```

type UpdateIpRequest struct {
	Zone utils.Zone

	IpId string

	Reverse *string
	Server  *string
	Override  map[string]interface{}
}
func (this *UpdateIpRequest) MarshalJSON() ([]byte, error) {
	if this.Override==nil {
		this.Override=make(map[string]interface{})
	}
       if this.Reverse!=nil { this.Override["reverse"]=this.Reverse}
       if this.Server!=nil { this.Override["server"]=this.Server}
	return json.Marshal(this.Override)
}
```
For detach IP:
```
_,err:=instanceApi.UpdateIp(&instance.UpdateIpRequest{	
	Zone: utils.ZoneFrPar1,
	IpId:"1511-1515-5858-525",
	Override:map[string]interface{}{"server":nil},
})
```

What do you think ?

kindermoumoute:
fix #12 
fix #25

### DIFF
--- a/api/instance/v1/instance_sdk.go
+++ b/api/instance/v1/instance_sdk.go
@@ -34,14 +34,14 @@ var (
 	_ = parameter.AddToQuery
 )
 
-// Api: instance API
-type Api struct {
+// API: instance API
+type API struct {
 	client *scw.Client
 }
 
-// NewApi returns a Api object from a Scaleway client.
-func NewApi(client *scw.Client) *Api {
-	return &Api{
+// NewAPI returns a API object from a Scaleway client.
+func NewAPI(client *scw.Client) *API {
+	return &API{
 		client: client,
 	}
 }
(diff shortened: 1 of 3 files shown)


kindermoumoute:
see #44

PR #46, opened 2019-05-19
feat: add pagination support

jerome-quere:


scw/request.go
```diff
@@ -66,17 +74,47 @@ func (c *Client) Do(req *ScalewayRequest, opts ...RequestOption) (*http.Response
...
+	for i := uint64(0); i < total; i++ {
+		r := reflect.New(reflect.ValueOf(res).Type())
+		req.Pagination.Page = 1      // Init pagination
+		err := c.Do(req, r, opts...) // Do first request
+		if err != nil {
+			return err
+		}
+		slice = append(slice, reflect.ValueOf(res).FieldByName("????").Interface().([]interface{})...)
+	}
+	reflect.ValueOf(res).FieldByName("????").Set(reflect.ValueOf(slice))
```
kindermoumoute: This is where and interface would be powerful:
```Go
type lister interface {
    Results() []interface{}
    Append([]interface{})
}
//...

switch v := res.(type) {
case lister:
	// do first request to get total
	
	var results []interface{}
	req.Pagination.Page = 0      // Init pagination
	for i := uint64(0); i < total; i+=len(results) {
		r := reflect.New(reflect.ValueOf(res).Type())
		req.Pagination.Page++
		err := c.Do(req, r, opts...) // Do first request
		if err != nil {
			return err
		}
		results = r.(lister).Results()
		v.Append(results)
	}
default:
    return fmt.Errorf("pagination is not supported on %s", v)
}
```

scw/client.go
```diff
@@ -90,6 +91,10 @@ func defaultOptions() []ClientOption {
 	return []ClientOption{
 		WithAPIURL("https://api.scaleway.com"),
 		WithUserAgent(userAgent),
+		WithPaginationConfig(&PaginationConfig{
```
kindermoumoute: IMO clientOption must have static parameters:

```Go
WithPaginationConfig(full bool, pageSize uint)
```

Let's decide fully decide what this client option is about so we fix these parameters for good.

scw/client_option.go
```diff
@@ -121,3 +121,14 @@ func WithDefaultZone(zone utils.Zone) ClientOption {
 		s.defaultZone = zone
 	}
 }
+
+func WithPaginationConfig(paginationConfig *PaginationConfig) ClientOption {
```
kindermoumoute: `WithPagination()`

scw/client.go
```diff
@@ -90,6 +91,10 @@ func defaultOptions() []ClientOption {
...
+			Full:     false,
```
QuentinBrosse: What is the exact purpose of `full`? If it's true the pagination is disabled?
Maybe we could find a better name

scw/client_option.go
```diff
@@ -121,3 +121,14 @@ func WithDefaultZone(zone utils.Zone) ClientOption {
...
+	return func(s *settings) {
+		s.paginationConfig = *paginationConfig
+	}
+}
+
+func WithoutPagination() ClientOption {
```
kindermoumoute: We should achieve "no pagination" (what does it mean?) with `WithPagination`

PR #55, opened 2019-05-22
feat: implement client.WithDefaultPageSize

QuentinBrosse:
Blocked by: #57

### DIFF
--- a/scw/client.go
+++ b/scw/client.go
@@ -101,7 +101,7 @@ func (c *Client) GetDefaultZone() (utils.Zone, bool) {
 }
 
 // GetDefaultPageSize return the default page size of the client.
-// This value can be set from the client option
+// This value can be set in the client option
 // WithDefaultPageSize(). Be aware this value can be empty.
 func (c *Client) GetDefaultPageSize() (int32, bool) {
 	return *c.defaultPageSize, c.defaultPageSize != nil
--- a/scw/client.go
+++ b/scw/client.go
@@ -26,6 +26,7 @@ type Client struct {
 	defaultOrganizationID *string
 	defaultRegion         *utils.Region
 	defaultZone           *utils.Zone
+	defaultPageSize       *int32
 }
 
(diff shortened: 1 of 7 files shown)


scw/client_option.go
```diff
@@ -121,3 +121,12 @@ func WithDefaultZone(zone utils.Zone) ClientOption {
 		s.defaultZone = zone
 	}
 }
+
+// WithDefaultPageSize client option overrides the default page size of the SDK.
+//
+// It will be used as the default value of the page_size field in all requests made with this client.
+func WithDefaultPageSize(pageSize int32) ClientOption {
+	return func(s *settings) {
+		s.defaultPageSize = pageSize
```
jerome-quere: should check that page size is not 0

scw/client.go
```diff
@@ -89,6 +91,13 @@ func (c *Client) GetDefaultZone() utils.Zone {
 	return c.defaultZone
 }
 
+// GetDefaultPageSize return the default page size of the client.
+// This value can be set from the client option
```
DenBeke: can be set **in** the client option

scw/client_option.go
kindermoumoute: The check should not happen here but in [`settings.validate()`](https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/settings.go#L33)

scw/client.go
```diff
@@ -98,6 +100,13 @@ func (c *Client) GetDefaultZone() (utils.Zone, bool) {
 	return utils.Zone(""), false
 }
 
+// GetDefaultPageSize return the default page size of the client.
+// This value can be set in the client option
+// WithDefaultPageSize(). Be aware this value can be empty.
+func (c *Client) GetDefaultPageSize() (int32, bool) {
+	return *c.defaultPageSize, c.defaultPageSize != nil
```
kindermoumoute: will panic if c.defaultPageSize is nil

PR #56, opened 2019-05-22
feat: list all pages

kindermoumoute:


scw/lister.go
```diff
@@ -0,0 +1,54 @@
+package scw
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+)
+
+type lister interface {
```
jerome-quere: this is not the interface we talk about.

scw/lister.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// doListAll collect all pages of a List request and aggregate all results on a single response.
+func (c *Client) doListAll(req *ScalewayRequest, res interface{}) (err error) {
+	switch response := res.(type) {
+	case lister:
+		n := 0
+		page := 1 // start at page 1
+		totalCount := -1
+		for i := 0; i < totalCount; i += n {
```
jerome-quere: will never enter the loop

scw/lister.go
```diff
@@ -0,0 +1,54 @@
...
+			// set current page
+			req.Query.Set("page", strconv.Itoa(page))
```
jerome-quere: We should avoid mutate the request (make a copy ?)

api/lb/v1/lb_sdk.go
```diff
@@ -707,6 +707,20 @@ type ListLbsResponse struct {
 	TotalCount uint32 `json:"total_count,omitempty"`
 }
 
+func (r *ListLbsResponse) GetTotalCount() int {
```
DenBeke: Is this generated code?

kindermoumoute: Soon-to-be

DenBeke: > Soon-to-be

👌

Just don't forget the Godoc 😜

scw/lister.go
kindermoumoute: I can copy the query parameters at the beginning and set it back at the end, does it work for you?

scw/client.go
```diff
@@ -110,16 +117,35 @@ func (c *Client) GetDefaultPageSize() (int32, bool) {
 	return 0, false
 }
 
-// Do performs an HTTP request based on the ScalewayRequest object.
-// RequestOptions are executed on the ScalewayRequest.
+// Do perform HTTP request(s) based on the ScalewayRequest object.
```
DenBeke: perform**s**

DenBeke:
LGTM

PR #57, opened 2019-05-22
refactor: client getters

QuentinBrosse:
**This changes will break the public API.**

This PR refactor clients options getters. From now, they will return their values and a boolean that precise if the value has been set or not (with a `scw.WithDefaultSomething()`).

```golang
--- func (c *Client) GetDefaultOrganizationID() string
+++ func (c *Client) GetDefaultOrganizationID() (string, bool)

--- func (c *Client) GetDefaultRegion() utils.Region
+++ func (c *Client) GetDefaultRegion() (utils.Region, bool)

--- func (c *Client) GetDefaultZone() utils.Zone
+++ func (c *Client) GetDefaultZone() (utils.Zone, bool)
```

### DIFF
--- a/scw/client.go
+++ b/scw/client.go
@@ -23,9 +23,9 @@ type Client struct {
 	auth                  auth.Auth
 	apiURL                string
 	userAgent             string
-	defaultOrganizationID string
-	defaultRegion         utils.Region
-	defaultZone           utils.Zone
+	defaultOrganizationID *string
+	defaultRegion         *utils.Region
+	defaultZone           *utils.Zone
 }
 
 // NewClient instantiates a new Client object.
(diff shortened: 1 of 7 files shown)


scw/client.go
```diff
@@ -69,24 +69,24 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 }
 
 // GetDefaultOrganizationID return the default organization ID
-// of the client. This value can be set from the client option
+// of the client. This value can be set in the client option
 // WithDefaultOrganizationID(). Be aware this value can be empty.
-func (c *Client) GetDefaultOrganizationID() string {
-	return c.defaultOrganizationID
+func (c *Client) GetDefaultOrganizationID() (string, bool) {
+	return *c.defaultOrganizationID, c.defaultOrganizationID != nil
```
kindermoumoute: will panic if c.defaultOrganizationID is nil

scw/client.go
```diff
@@ -69,24 +69,24 @@ func NewClient(opts ...ClientOption) (*Client, error) {
...
 }
 
 // GetDefaultRegion return the default region of the client.
-// This value can be set from the client option
+// This value can be set in the client option
 // WithDefaultRegion(). Be aware this value can be empty.
-func (c *Client) GetDefaultRegion() utils.Region {
-	return c.defaultRegion
+func (c *Client) GetDefaultRegion() (utils.Region, bool) {
+	return *c.defaultRegion, c.defaultRegion != nil
```
kindermoumoute: will panic if c.defaultRegion is nil

scw/client.go
```diff
@@ -69,24 +69,24 @@ func NewClient(opts ...ClientOption) (*Client, error) {
...
 }
 
 // GetDefaultZone return the default zone of the client.
-// This value can be set from the client option
+// This value can be set in the client option
 // WithDefaultZone(). Be aware this value can be empty.
-func (c *Client) GetDefaultZone() utils.Zone {
-	return c.defaultZone
+func (c *Client) GetDefaultZone() (utils.Zone, bool) {
+	return *c.defaultZone, c.defaultZone != nil
```
kindermoumoute: will panic if c.defaultZone is nil

Issue #34, opened 2019-05-16
Cannot detach volume from instance


kindermoumoute:
When trying to detach a volume with `UpdateServer`, a nil volume results in the following error:

>" type is invalid_request_error: No saved boot_type to restore"

cc @Archie1978

PR #58, opened 2019-05-22
feat: instance attach/detach volume

DenBeke:


api/instance/v1/instance_utils.go
```diff
@@ -86,3 +88,100 @@ func (s *API) UpdateIP(req *UpdateIPRequest, opts ...scw.RequestOption) (*Update
...
 }
+
+// AttachVolumeRequest contains the parameters to attach a volume to a server
+type AttachVolumeRequest struct {
+	Zone     utils.Zone `json:"-"`
+	ServerID string     `json:"-"`
+	VolumeID string     `json:"-"`
+}
+
+// AttachVolumeResponse [TODO]
```
kindermoumoute: TODO

api/instance/v1/instance_utils.go
```diff
@@ -86,3 +88,100 @@ func (s *API) UpdateIP(req *UpdateIPRequest, opts ...scw.RequestOption) (*Update
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	volumes := getServerResponse.Server.Volumes
+
+	newVolumes := volumesToVolumeTemplates(volumes)
+
+	// add volume to volumes list
+	newVolumes[fmt.Sprintf("%d", len(volumes))] = &VolumeTemplate{ID: req.VolumeID, Name: req.VolumeID /* name is ignored on this PATCH */}
```
QuentinBrosse:
```golang
key := fmt.Sprintf("%d", len(volumes))
newVolumes[key] = &VolumeTemplate{
    ID: req.VolumeID,
    // name is ignored on this PATCH
    Name: req.VolumeID
}
```

api/instance/v1/instance_utils.go
```diff
@@ -86,3 +88,100 @@ func (s *API) UpdateIP(req *UpdateIPRequest, opts ...scw.RequestOption) (*Update
...
+		return nil, err
+	}
+
+	return &AttachVolumeResponse{Server: updateServerResponse.Server}, nil
+}
+
+// DetachVolumeRequest contains the parameters to detach a volume from a server
+type DetachVolumeRequest struct {
+	Zone     utils.Zone `json:"-"`
+	ServerID string     `json:"-"`
```
jerome-quere: Do we need server id ? Can this be get with /volume/id ?

api/instance/v1/instance_utils_test.go
```diff
@@ -28,93 +28,127 @@ func TestAttachIP(t *testing.T) {
 		reverse      = "1.1.1.1"
 	)
 
-	// Create server
-	createServerResponse, err := instanceAPI.CreateServer(&CreateServerRequest{
-		Zone:         zone,
-		Name:         "instance_utils_test",
-		Organization: organization,
-		Image:        image,
+	t.Run("create server", func(t *testing.T) {
```
jerome-quere: why is this test in `TestAttachIP` ?

DenBeke: I reused added it to the test scenario (to reuse setup/teardown). I should be change the name of the function.

api/instance/v1/instance_utils.go
DenBeke: If we do one more call we could omit the `ServerID`.

jerome-quere: Let's do one more call

DenBeke: 👍

api/instance/v1/instance_utils.go
```diff
@@ -86,3 +88,117 @@ func (s *API) UpdateIP(req *UpdateIPRequest, opts ...scw.RequestOption) (*Update
...
+	// get volume
+	getVolumeResponse, err := s.GetVolume(&GetVolumeRequest{
+		Zone:     req.Zone,
+		VolumeID: req.VolumeID,
+	})
+	if err != nil {
+		return nil, err
+	}
+	if getVolumeResponse.Volume == nil || getVolumeResponse.Volume.Server == nil {
+		return nil, fmt.Errorf("expected volume to have value in response")
```
jerome-quere: we should slip in to different error.
// expected volume to have value in response

and 
// server should be attach to a server

jerome-quere: if if getVolumeResponse.Volume == nil {
return nil, fmt.Errorf("expected volume to have value in response")
}

if (getVolumeResponse.Volume.Server == nil) {
return nil, fmt.Errorf("server should be attach to a server")
}

jerome-quere:
fix #34

PR #59, opened 2019-05-23
fix: fixed error check on `client.Do`

DenBeke:


scw/client.go
```diff
@@ -155,7 +155,10 @@ func (c *Client) Do(req *ScalewayRequest, res interface{}, opts ...RequestOption
 	}
 
 	defer func() {
-		err = resp.Body.Close()
+		tmpErr := resp.Body.Close()
+		if tmpErr != nil {
```
QuentinBrosse: If there is a previous error and a `Body.Close()` error, we have to prioritize the previous one. 

```golang
if tmp == nil && tmpErr != nil { 
```

PR #60, opened 2019-05-23
test: enforce x-auth-token removal

DenBeke:


internal/testhelpers/httprecorder/check_tokens.sh
```diff
@@ -0,0 +1,14 @@
...
+for file in $files
+do
+    echo "checking $file";
+    if grep -i --quiet "x-auth-token" $file
+    then
+        echo "found x-auth-token in file $file";
+        exit 1;
+    fi
+done
+exit 0;
```
QuentinBrosse: add newline

internal/testhelpers/httprecorder/check_tokens.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
QuentinBrosse: You don't think that it's a little bit strange to have a `shell` script in a go package? :/

internal/testhelpers/httprecorder/recorder.go
```diff
@@ -41,6 +53,15 @@ func CreateRecordedScwClient(cassetteName string) (*scw.Client, *recorder.Record
 	// Add a filter which removes Authorization headers from all requests:
 	r.AddFilter(func(i *cassette.Interaction) error {
 		delete(i.Request.Headers, "x-auth-token")
+		delete(i.Request.Headers, "X-Auth-Token")
+
+		if UpdateCassette {
+			secretKey, _ := config.GetSecretKey()
+			if i != nil && strings.Contains(fmt.Sprintf("%v", *i), secretKey) {
+				panic(errors.New("found secret key in cassette"))
```
QuentinBrosse: `fmt.Errorf` (for consistency)?

internal/testhelpers/httprecorder/check_tokens.sh
DenBeke: It depends on the way you look at it. Having a script in a Go package, next to the files it is related to doesn't seem strange to me...

But tell me which other location you prefer.

internal/testhelpers/httprecorder/recorder.go
DenBeke: yeps

PR #62, opened 2019-05-24
fix: double body close on hasResponseError

jerome-quere:


scw/client.go
```diff
@@ -242,3 +242,25 @@ func setInsecureMode(c httpClient) {
...
+		return newErr
+	}
+
+	err := json.NewDecoder(res.Body).Decode(newErr)
+	if err != nil {
+		return err
+	}
+
+	return newErr
+}
```
DenBeke: newline

DenBeke:
LGTM

PR #67, opened 2019-05-31
feat: debug http response

kindermoumoute:


scw/client.go
```diff
@@ -198,6 +198,14 @@ func (c *Client) do(req *ScalewayRequest, res interface{}) (err error) {
 			err = closeErr
 		}
 	}()
+	if logger.ShouldLog(logger.LogLevelDebug) {
+		dump, err := httputil.DumpResponse(resp, true)
```
DenBeke: Can this contain sensitive data?

jerome-quere: in response hopfully not.

kindermoumoute: Nope

DenBeke: 👍

PR #79, opened 2019-06-05
feat: remove error from Parse{Zone|Region} signatures

QuentinBrosse:


DenBeke:
IIRC the (currently) unused error was added to make it possible to do some error checking in the feature without doing a breaking change.

QuentinBrosse:
@DenBeke IMH we need to support custom `zone` / `region` so the warning is enough. `{Zone|Region}.Exists()` is here if we need to check the existence.

PR #80, opened 2019-06-05
feat: add instance.WaitForServer

jerome-quere:
feat: add WaitForServer


jerome-quere:
Should we rename to `WaitServer` ?

DenBeke:
> Should we rename to `WaitServer` ?

I prefer `WaitForServer`, @jerome-quere

PR #81, opened 2019-06-05
feat: add injector

jerome-quere:
This will allow us to inject data like organization / pageSize / perPage in a generic way.

### DIFF
--- a/internal/injector/injector.go
+++ b/internal/injector/injector.go
@@ -1,0 +1,47 @@
+package injector
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// Inject value inside a struct based on struct tags
+func Inject(target interface{}, tagName string, data map[string]interface{}) error {
+	targetValue := reflect.ValueOf(target)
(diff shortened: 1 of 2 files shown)


internal/injector/injector.go
```diff
@@ -0,0 +1,62 @@
...
+		fieldValue := targetValue.Field(i)
+		fieldType := fieldValue.Type()
+
+		tagValue, tagExist := field.Tag.Lookup(tagName)
+		if !tagExist {
+			continue
+		}
+
+		// Handle json tag format (e.g. json:"name,omitempty")
+		tagValue = strings.SplitN(tagValue, ",", 2)[0]
```
kindermoumoute: one day this will segfault  🎩

internal/injector/injector_test.go
```diff
@@ -0,0 +1,128 @@
+package injector
```
kindermoumoute: It would be interesting to add tests with actual generated struct. These would probably live in `scw` package

PR #92, opened 2019-06-17
refactor: instance metadata methods return error

jerome-quere:


api/instance/v1/instance_metadata_sdk.go
```diff
@@ -197,7 +196,7 @@ func (*MetadataAPI) GetUserdata(key string) ([]byte, scw.SdkError) {
 }
 
 // SetUserdata sets the userdata key with the value value
```
kindermoumoute: with the given value*

PR #96, opened 2019-06-18
feat: do action and wait

kindermoumoute:


api/instance/v1/server_utils.go
```diff
@@ -51,3 +52,51 @@ func (s *API) WaitForServer(req *WaitForServerRequest) (*Server, error) {
 
 	return server.(*Server), err
 }
+
+// DoActionAndWaitRequest is used by DoActionAndWait method
+type DoActionAndWaitRequest struct {
+	ServerID string
+	Zone     utils.Zone
+	Action   ServerAction
+	Timeout  time.Duration
```
jerome-quere: Add comment with default value  and add a default value

api/instance/v1/server_utils.go
```diff
@@ -51,3 +52,51 @@ func (s *API) WaitForServer(req *WaitForServerRequest) (*Server, error) {
...
+}
+
+// DoActionAndWait start an action and wait for the server to be in the correct "terminal state"
+// expected by this action.
+func (s *API) DoActionAndWait(req *DoActionAndWaitRequest) error {
```
jerome-quere: If underlying method is `ServerAction` should this be `ServerActionAndWait` ?
or maybe rename underlying action

api/instance/v1/server_utils.go
```diff
@@ -51,3 +52,51 @@ func (s *API) WaitForServer(req *WaitForServerRequest) (*Server, error) {
...
+	// check the action was properly executed
+	expectedState := ServerState("")
+	switch req.Action {
+	case ServerActionPoweron, ServerActionReboot:
+		expectedState = ServerStateRunning
+	case ServerActionPoweroff:
+		expectedState = ServerStateStopped
+	case ServerActionStopInPlace:
+		expectedState = ServerStateStoppedInPlace
+	}
```
jerome-quere: Add a default `expectedState := ServerState("") `

api/instance/v1/server_utils.go
```diff
@@ -51,3 +52,51 @@ func (s *API) WaitForServer(req *WaitForServerRequest) (*Server, error) {
...
+
+	// backup can be performed from any state
+	if req.Action != ServerActionBackup && finalServer.State != expectedState {
```
jerome-quere: replace  `req.Action != ServerActionBackup` by `expectedState != ServerState("")`

work well with switch default (see  previous comment)

api/instance/v1/server_utils.go
jerome-quere: I don't like to rely on someone else default value  (in this case `waitForServer`) in a method

api/instance/v1/server_utils.go
```diff
@@ -4,25 +4,20 @@ import (
...
 
-// WaitForServer wait for the server to be in a "terminal state" before returning.
+// waitForServer wait for the server to be in a "terminal state" before returning.
 // This function can be used to wait for a server to be started for example.
-func (s *API) WaitForServer(req *WaitForServerRequest) (*Server, error) {
-
-	if req.Timeout == 0 {
-		req.Timeout = 5 * time.Minute
-	}
+func (s *API) waitForServer(req *waitForServerRequest) (*Server, error) {
```
jerome-quere: now that this is private it should return a SdkError ?

kindermoumoute: done

PR #101, opened 2019-06-18
fix: seed namegenerator and rename package

QuentinBrosse:


namegenerator/name_generator.go
```diff
@@ -1,12 +1,17 @@
...
+package namegenerator
 
 import (
 	"math/rand"
 	"strings"
+	"time"
 )
 
+func init() {
+	rand.Seed(time.Now().UnixNano())
```
jerome-quere: Please use https://golang.org/pkg/math/rand/#NewSource

PR #105, opened 2019-06-18
feat: update generated apis

jerome-quere:


api/instance/v1/instance_sdk.go
```diff
@@ -2600,7 +2592,9 @@ type CreateSecurityGroupRequest struct {
 
 	Name string `json:"name,omitempty"`
 
-	OrganizationKey string `json:"organization_key,omitempty"`
+	Description string `json:"description,omitempty"`
+
+	Organization string `json:"organization,omitempty"`
```
kindermoumoute: was this change intended?

DenBeke: Apparently @jerome-quere changed it yesterday due to an error in the proto.

api/lb/v1/lb_sdk.go
```diff
@@ -1542,45 +1542,45 @@ type UpdateBackendRequest struct {
 	// BackendID backend ID to update
```
kindermoumoute: why this request remain public?

DenBeke: Why would we make it private? 🤔

PR #106, opened 2019-06-19
feat: add instance.UpdateSecurityGroup

jerome-quere:


api/instance/v1/security_group_utils_test.go
```diff
@@ -0,0 +1,57 @@
...
+		OutboundDefaultPolicy: &accept,
+	})
+
+	testhelpers.Ok(t, err)
+	testhelpers.Equals(t, "new_name", updateResponse.SecurityGroup.Name)
+	testhelpers.Equals(t, "new_description", updateResponse.SecurityGroup.Description)
+	testhelpers.Equals(t, SecurityGroupPolicyDrop, updateResponse.SecurityGroup.InboundDefaultPolicy)
+	testhelpers.Equals(t, SecurityGroupPolicyAccept, updateResponse.SecurityGroup.OutboundDefaultPolicy)
+	testhelpers.Equals(t, false, updateResponse.SecurityGroup.Stateful)
+}
```
DenBeke: You should also delete the resource you create. Otherwise we'll have dangling resources when updating the cassettes.

api/instance/v1/security_group_utils.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"fmt"
+
+	"github.com/scaleway/scaleway-sdk-go/internal/errors"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/scaleway-sdk-go/utils"
+)
+
+// UpdateSecurityGroupRequest contains the parameters to update a security group
+type UpdateSecurityGroupRequest struct {
```
QuentinBrosse: This is the public struct, we should add godoc on it

PR #110, opened 2019-06-20
fix: wait for deleted instance on terminate

kindermoumoute:


Sh4d1:
Is it an option? Or does this PR make the call to delete instance blocking?

PR #111, opened 2019-06-20
fix: scw.request.getAllHeaders

QuentinBrosse:


scw/request.go
```diff
@@ -34,11 +34,12 @@ func (req *ScalewayRequest) getAllHeaders(token auth.Auth, userAgent string, ano
 
 	allHeaders.Set("User-Agent", userAgent)
 	if req.Body != nil {
-		allHeaders.Set("content-type", "application/json")
+		allHeaders.Set("Content-Type", "application/json")
```
DenBeke: FYI: this doesn't change the outcome since `Set` already canonicalizes the headers.

> Set sets the header entries associated with key to the single element value. It replaces any existing values associated with key. The key is case insensitive; it is canonicalized by textproto.CanonicalMIMEHeaderKey. To use non-canonical keys, assign to the map directly.

QuentinBrosse: Thanks for the info! :) Just for consistency, we could keep this version. :)

DenBeke: Yeah, I completely agree to keep this version! 😊

scw/request.go
```diff
@@ -34,11 +34,12 @@ func (req *ScalewayRequest) getAllHeaders(token auth.Auth, userAgent string, ano
 
 	allHeaders.Set("User-Agent", userAgent)
 	if req.Body != nil {
-		allHeaders.Set("content-type", "application/json")
```
kindermoumoute: FYI: you don't need to change this line because the key is formatted https://golang.org/src/net/textproto/reader.go?s=15268:15312#L563

PR #112, opened 2019-06-20
feat: support raw body requests

QuentinBrosse:


scw/client.go
```diff
@@ -232,9 +233,24 @@ func (c *Client) do(req *ScalewayRequest, res interface{}) (sdkErr SdkError) {
...
+
+		switch contentType {
+		case "application/json":
+			err = json.NewDecoder(httpResponse.Body).Decode(&res)
+			if err != nil {
+				return errors.Wrap(err, "could not parse %s response body", contentType)
+			}
+		default:
+			buffer, ok := res.(io.Writer)
+			if ok {
```
jerome-quere: nit: `if buffer, isBuffer := res.(io.Writer) ; isBuffer {`

QuentinBrosse:
@DenBeke 
> Do we test the do somewhere?

Not yet

PR #119, opened 2019-06-28
feat: lock server patch

kindermoumoute:


api/instance/v1/server_utils.go
```diff
@@ -13,6 +13,14 @@ import (
 	"github.com/scaleway/scaleway-sdk-go/utils"
 )
 
+type UpdateServerRequest updateServerRequest
+
+// updateServer update server
+func (s *API) UpdateServer(req *UpdateServerRequest, opts ...scw.RequestOption) (*UpdateServerResponse, error) {
+	defer lockServer(req.Zone, req.ServerID).Unlock()
+	return s.updateServer((*updateServerRequest)(req), opts...)
```
DenBeke: Maybe I'm overlooking something, but where do you lock the server resource?

kindermoumoute: in `lockServer` 👍

DenBeke: But `lockServer` is executed in the defer?

kindermoumoute: Onyl the last function is deferred

jerome-quere: Unlock is defered not lock

DenBeke: TIL...

PR #129, opened 2019-07-04
feat: update generated apis

jerome-quere:


api/instance/v1/instance_sdk.go
```diff
@@ -2975,13 +2975,13 @@ type setSecurityGroupRuleRequest struct {
...
 
-	DestPortFrom uint32 `json:"dest_port_from"`
-
-	DestPortTo uint32 `json:"dest_port_to"`
-
 	Position uint32 `json:"position"`
 
 	Editable bool `json:"editable"`
+
+	DestPortFrom **uint32 `json:"dest_port_from"`
```
kindermoumoute: wait, this field does not have `omitempty` ==> one pointer should be enough 🤔

PR #130, opened 2019-07-04
fix: handle resource in error

kindermoumoute:


scw/errors.go
```diff
@@ -44,3 +55,25 @@ func (e *ResponseError) Error() string {
 
 // IsScwSdkError implement SdkError interface
 func (e *ResponseError) IsScwSdkError() {}
+
+func hasResponseError(res *http.Response) SdkError {
```
QuentinBrosse: godoc

scw/errors.go
```diff
@@ -18,6 +22,9 @@ type ResponseError struct {
 	// Type is a string code that defines the kind of error
 	Type string `json:"type,omitempty"`
 
+	// Resource is a string code that defines the resource concerned by the error
+	Resource string `json:"resource,omitempty"`
```
jerome-quere: Can you add a comment that this only concert instance API

PR #131, opened 2019-07-05
feat: add NullableStringValue

kindermoumoute:


api/instance/v1/instance_utils.go
```diff
@@ -303,3 +295,24 @@ func (r *ListServersTypesResponse) UnsafeAppend(res interface{}) (int, scw.SdkEr
...
+	if string(b) == "null" {
+		v.Null = true
+		return nil
+	}
+
+	var tmp string
+	if err := json.Unmarshal(b, &tmp); err != nil {
+		return err
+	}
+	v.Value = tmp
```
jerome-quere: You should add `v.Null = false` as someone could reuse an already used object while Unmarshaling

kindermoumoute: Good catch!

PR #139, opened 2019-07-30
feat: implement scw.Size type

QuentinBrosse:


scw/size.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	"github.com/dustin/go-humanize"
+)
+
+type Size uint64
+
+const (
+	Byte Size = 1 << (iota * 10)
+	KiByte
```
jerome-quere: maybe KByte is easier

scw/size.go
```diff
@@ -0,0 +1,21 @@
+package scw
+
+import (
+	"fmt"
+)
+
+type Size uint64
```
kindermoumoute: move to custome_type.go

vendor/modules.txt
```diff
@@ -1,5 +1,5 @@
 # github.com/dnaeon/go-vcr v1.0.1
-github.com/dnaeon/go-vcr/recorder
 github.com/dnaeon/go-vcr/cassette
+github.com/dnaeon/go-vcr/recorder
```
kindermoumoute: can we avoid this change ?

PR #149, opened 2019-08-08
feat: auth request option

kindermoumoute:


scw/client.go
```diff
@@ -124,22 +124,20 @@ func (c *Client) GetDefaultPageSize() (int32, bool) {
 // Do performs HTTP request(s) based on the ScalewayRequest object.
 // RequestOptions are applied prior to doing the request.
 func (c *Client) Do(req *ScalewayRequest, res interface{}, opts ...RequestOption) (err error) {
-	requestSettings := newRequestSettings()
```
QuentinBrosse: This line was useful to avoid mutations on the client request object.

PR #158, opened 2019-08-23
fix: use standard errors

kindermoumoute:


internal/async/wait.go
```diff
@@ -84,6 +85,6 @@ func WaitSync(config *WaitSyncConfig) (terminalValue interface{}, err error) {
 		return nil, err
 	case <-time.After(config.Timeout):
 		timeout <- true
-		return nil, fmt.Errorf("timeout after %v", config.Timeout)
+		return nil, errors.New("timeout after %v", config.Timeout)
```
jerome-quere: We say that we should return internal error only on public method to avoid double wrapping.

internal/errors/error.go
```diff
@@ -4,20 +4,20 @@ import "fmt"
 
 // Error is a base error that implement scw.SdkError
 type Error struct {
-	str string
-	err error
+	Str string
+	Err error
```
jerome-quere: Why make it public ? this is why `Unwrap` is for
Same question for `Str`

kindermoumoute: This is an internal package. It will allow to add error helpers in `scw`.

PR #165, opened 2019-09-02
feat: rename project id to organization id

QuentinBrosse:


QuentinBrosse:
@kindermoumoute 
> ```
> api/test/v1/test_sdk.go:        ProjectID string `json:"project_id"`
> api/test/v1/test_sdk.go:        parameter.AddToQuery(query, "project_id", req.ProjectID)
> api/test/v1/test_sdk.go:        ProjectID string `json:"project_id"`
> ```

I intentionally keep `project_id` for this API. We need to update the API itself before making changes in the SDK.

PR #166, opened 2019-09-02
doc: add create server exemple

jerome-quere:


example_test.go
```diff
@@ -84,6 +87,68 @@ func Example_listServers() {
 	fmt.Println(response)
 }
 
+func Example_createServer() {
```
jerome-quere: 60+ line to create a server 😞

example_test.go
```diff
@@ -84,6 +87,68 @@ func Example_listServers() {
...
+	createIPRes, err := instanceAPI.CreateIP(&instance.CreateIPRequest{})
+	if err != nil {
+		panic(err)
+	}
+
+	// Create a new DEV1-S server
+	createRes, err := instanceAPI.CreateServer(&instance.CreateServerRequest{
+		Name:              "my-server-01",
+		CommercialType:    "DEV1-S",
+		Image:             imageID,
```
jerome-quere: Would be nice if I could have write `ubuntu-bionic` here

example_test.go
```diff
@@ -84,6 +87,68 @@ func Example_listServers() {
...
+		DynamicIPRequired: scw.BoolPtr(false),
```
jerome-quere: Sad that this line is required if you want to use a dynamic IP.

example_test.go
jerome-quere: We should consider a `Create(Simple|Basic)Server` function

example_test.go
```diff
@@ -84,6 +87,68 @@ func Example_listServers() {
...
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// Start the server and wait until it's ready.
+	err = instanceAPI.ServerActionAndWait(&instance.ServerActionAndWaitRequest{
+		ServerID: createRes.Server.ID,
+		Action:   instance.ServerActionPoweron,
+		Timeout:  3 * time.Minute,
```
kindermoumoute: 10 minutes

example_test.go
```diff
@@ -84,6 +87,68 @@ func Example_listServers() {
...
+	}
+
+	// Create SDK objects for Scaleway Instance and marketplace
+	instanceAPI := instance.NewAPI(client)
+	marketplaceAPI := marketplace.NewAPI(client)
+
+	// Get imageID for ubuntu-bionic from marketplace API
+	imageID, err := marketplaceAPI.GetLocalImageIDByLabel(&marketplace.GetLocalImageIDByLabelRequest{
+		ImageLabel:     "ubuntu-bionic",
+		CommercialType: "DEV1-S",
```
kindermoumoute: nit: put `"DEV1-S"` and `"ubuntu-bionic"` in a const at the top

PR #177, opened 2019-09-11
feat: standard error structures

kindermoumoute:


scw/errors.go
```diff
@@ -75,6 +77,55 @@ func hasResponseError(res *http.Response) SdkError {
...
 	}
+	stdErr := decodeStandardError(newErr.Type, res.Body)
+	if stdErr != nil {
+		return stdErr
+	}
 
 	return newErr
 }
+
+func decodeStandardError(errorType string, body io.ReadCloser) SdkError {
```
jerome-quere: unmarshalStandardError

scw/errors.go
```diff
@@ -75,6 +77,55 @@ func hasResponseError(res *http.Response) SdkError {
...
+	switch errorType {
+	case "invalid_arguments":
```
jerome-quere: avoid code duplication by moving code after switch

scw/errors.go
```diff
@@ -75,6 +77,55 @@ func hasResponseError(res *http.Response) SdkError {
 	if err != nil {
 		return errors.Wrap(err, "could not parse error response body")
 	}
+	stdErr := decodeStandardError(newErr.Type, res.Body)
```
jerome-quere: res.Body is already consumed

PR #179, opened 2019-09-11
fix: UpdateSecurityGroupRule can now remove DestPortFrom and DestPortTo

jerome-quere:


api/instance/v1/security_group_utils.go
```diff
@@ -97,13 +97,16 @@ type UpdateSecurityGroupRuleRequest struct {
...
-	DestPortFrom *uint32                     `json:"dest_port_from,omitempty"`
-	DestPortTo   *uint32                     `json:"dest_port_to,omitempty"`
-	Position     *uint32                     `json:"position,omitempty"`
+	Protocol  *SecurityGroupRuleProtocol  `json:"protocol"`
+	Direction *SecurityGroupRuleDirection `json:"direction"`
+	Action    *SecurityGroupRuleAction    `json:"action"`
+	IPRange   *string                     `json:"ip_range"`
+	Position  *uint32                     `json:"position,omitempty"`
+
+	// If set to 0 DestPortFrom will be remove
```
QuentinBrosse: 1. "If set to 0, DestPortFrom will be removed."
2. It could be interesting to document a bit more this field

api/instance/v1/security_group_utils.go
```diff
@@ -97,13 +97,16 @@ type UpdateSecurityGroupRuleRequest struct {
...
+	DestPortFrom *uint32 `json:"dest_port_from"`
+	// If set to 0 DestPortTo will be remove
```
QuentinBrosse: 1. "If set to 0, DestPortFrom will be removed."
2. It could be interesting to document a bit more this field

PR #184, opened 2019-09-16
fix: attach volume key choice

Sh4d1:
Currently, when attaching a volume, the sdk attach it to the ID that is the len of the volume list. Unfortunately, if the list is `0, 1, 2` and the volume `1` is removed, the list becomes `0, 2`. This leads to a newly attached volumes to be attached to the ID `2`, thus overriding the already existing volume. This fix this issue by looping over all the keys, and attaching to a non existent key.


api/instance/v1/instance_utils.go
```diff
@@ -139,11 +139,22 @@ func (s *API) AttachVolume(req *AttachVolumeRequest, opts ...scw.RequestOption)
...
+				// name is ignored on this PATCH
+				Name: req.VolumeID,
+			}
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		return nil, fmt.Errorf("Could not find key to attach volume %s", req.VolumeID)
```
QuentinBrosse: could*

api/instance/v1/instance_utils.go
```diff
@@ -139,11 +139,22 @@ func (s *API) AttachVolume(req *AttachVolumeRequest, opts ...scw.RequestOption)
 	newVolumes := volumesToVolumeTemplates(volumes)
 
 	// add volume to volumes list
-	key := fmt.Sprintf("%d", len(volumes))
-	newVolumes[key] = &VolumeTemplate{
-		ID: req.VolumeID,
-		// name is ignored on this PATCH
-		Name: req.VolumeID,
+	found := false
```
QuentinBrosse: can you add a short comment to explain this approach, please?

PR #188, opened 2019-09-18
feat: update generated apis

QuentinBrosse:


api/instance/v1/instance_sdk.go
```diff
@@ -700,6 +700,18 @@ type Dashboard struct {
 	IPsUnused uint32 `json:"ips_unused"`
 }
 
+type ExtraNetworks struct {
+	Address string `json:"address,omitempty"`
```
jerome-quere: Why is there omit empty in a response message ?

QuentinBrosse: Because `ExtraNetworks` is used in request and response messages. I will create an `ExtraNetworksTemplate`.

PR #199, opened 2019-10-03
fix: scw.File unmarshal

Sh4d1:

### DIFF
--- a/scw/custom_types.go
+++ b/scw/custom_types.go
@@ -1,9 +1,11 @@
 package scw
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/scaleway/scaleway-sdk-go/internal/errors"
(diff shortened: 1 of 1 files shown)


scw/custom_types.go
```diff
@@ -37,6 +39,56 @@ type File struct {
 	Content io.Reader `json:"content"`
 }
 
+func (file *File) UnmarshalJSON(b []byte) error {
+	f := make(map[string]interface{})
```
kindermoumoute: Should use:
```
type file struct {
	Name string `json:"name"`
	ContentType string `json:"content_type"`
	Content []byte `json:"content"`
}
```

https://play.golang.org/p/kd26fvbDder

scw/custom_types.go
```diff
@@ -37,6 +39,32 @@ type File struct {
...
 }
 
+type file struct {
+	Name        string `json:"name"`
+	ContentType string `json:"content_type"`
+	Content     []byte `json:"content"`
+}
+
+func (f *File) UnmarshalJSON(b []byte) error {
+	var tmpFile file
```
kindermoumoute: Move the struct in the function
```
var tmpFile struct {
	Name        string `json:"name"`
	ContentType string `json:"content_type"`
	Content     []byte `json:"content"`
}
```

jerome-quere:
Close in favor of #201

PR #214, opened 2019-10-16
feat: update generated apis

QuentinBrosse:


QuentinBrosse:
Need manual changes done in #216

PR #218, opened 2019-10-21
feat: update generated apis

kindermoumoute:


api/instance/v1/instance_sdk.go
```diff
@@ -80,135 +80,135 @@ func (enum *Arch) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-type ComputeClusterPolicyMode string
+type GetServerTypesAvailabilityResponseAvailability string
```
jerome-quere: Should we move Availability enum out of `GetServerTypesAvailabilityResponse`

PR #225, opened 2019-10-31
feat: add client.GetAuthToken

QuentinBrosse:


scw/client.go
```diff
@@ -81,10 +81,22 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 	}, nil
 }
 
+// GetAuthToken return the auth of the client.
```
kindermoumoute: returns

scw/client.go
```diff
@@ -81,10 +81,22 @@ func NewClient(opts ...ClientOption) (*Client, error) {
...
+// This value can be set in the client option WithAuth().
+// Be aware this value can be empty.
+func (c *Client) GetAuthToken() (token *auth.Token, exists bool) {
+	if c.auth != nil {
```
kindermoumoute: nit:
```go
switch v := c.auth(type){
  case *auth.Token:
    return v, true
}
return nil, false
```

PR #228, opened 2019-11-05
feat(core): add ClientCredentialError

QuentinBrosse:


scw/client_option.go
```diff
@@ -172,7 +172,16 @@ func (s *settings) apply(opts []ClientOption) {
 func (s *settings) validate() error {
 	var err error
 	if s.token == nil {
-		return errors.New("no credential option provided")
+		return &ClientCredentialError{Type: clientCredentialError_NoOption}
+	}
+
+	if token, isToken := s.token.(*auth.Token); isToken {
+		if token.AccessKey == "" {
+			return &ClientCredentialError{Type: clientCredentialError_EmptyAccessKey}
```
kindermoumoute: The only issue I can see here is people using terraform will see this error because their config was migrated from the v1 (no access_key included).

QuentinBrosse: We can get an `access_key` from a `secret_key` with the account API.

kindermoumoute: Yes, the next question is where should it be done?

IMO we could do that when creating the SDK Client, if no access Key is provided, let's pull it.

QuentinBrosse: If we do it, it will be outside the validation and in another MR.

scw/client_option_test.go
```diff
@@ -37,11 +37,20 @@ func TestClientOptions(t *testing.T) {
 			},
 		},
 		{
-			name: "Should throw an credential error",
+			name: "Should throw a access key error",
```
kindermoumoute: an access*

scw/errors.go
```diff
@@ -255,3 +255,21 @@ func (e *OutOfStockError) Error() string {
...
+type clientCredentialErrorType string
+
+const (
+	clientCredentialError_EmptyAccessKey = clientCredentialErrorType("access key cannot be empty")
+	clientCredentialError_EmptySecreyKey = clientCredentialErrorType("secret key cannot be empty")
+)
+
+// clientCredentialError indicates that credentials have been badly provided for the client creation.
+type ClientCredentialError struct {
+	Type clientCredentialErrorType
```
jerome-quere: I would not make Type public as it is of a private type

PR #229, opened 2019-11-06
feat: add strcase lib

kindermoumoute:
- ToPublicGoName: convert a field name to a valid public Go name
- ToPrivateGoName: convert a field name to a valid private Go name
- ToBashArg: convert a field name to a valid bash argument

### DIFF
--- a/strcase/bash_arg.go
+++ b/strcase/bash_arg.go
@@ -5,7 +5,10 @@ import "strings"
 // ToBashArg returns the Bash public name of the given string
 func ToBashArg(s string) string {
 	s = ToPublicGoName(s)
-	s = strings.ReplaceAll(s, "IPv", "Ipv")
+	for _, initialism := range customInitialisms {
+		// catch this kind of pattern: ExampleIDs ==> ExampleIds ==> example-ids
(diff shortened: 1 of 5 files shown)


strcase/bash_arg.go
```diff
@@ -0,0 +1,58 @@
+package strcase
+
+import "strings"
+
+// ToBashArg returns the Bash public name of the given string
```
QuentinBrosse: dot

strcase/goname.go
```diff
@@ -0,0 +1,188 @@
+package strcase
+
+import (
+	"strings"
+	"unicode"
+)
+
+// ToPrivateGoName returns the Go public name of the given string
```
QuentinBrosse: dot

strcase/goname.go
```diff
@@ -0,0 +1,188 @@
...
+func ToPublicGoName(s string) string {
+	return toGoName(TitleFirstWord(s))
+}
+
+// ToPrivateGoName returns the Go private name of the given string
```
QuentinBrosse: dot

PR #236, opened 2019-11-13
feat: IPNet type

kindermoumoute:


scw/custom_types.go
```diff
@@ -164,3 +166,36 @@ func (tsp *TimeSeriesPoint) UnmarshalJSON(b []byte) error {
 
 	return nil
 }
+
+// IPNet inherits net.IPNet and represents an IP network.
+type IPNet struct {
```
jerome-quere: should it be a type alias ?

kindermoumoute: This way the type inherits `net.IPNet` methods

scw/custom_types.go
```diff
@@ -164,3 +166,36 @@ func (tsp *TimeSeriesPoint) UnmarshalJSON(b []byte) error {
...
+}
+
+func (n *IPNet) UnmarshalJSON(b []byte) error {
+	var str *string
+
+	err := json.Unmarshal(b, &str)
+	if err != nil {
+		return err
+	}
+	if str == nil {
```
jerome-quere: I am not sure this can be happen

scw/custom_types.go
```diff
@@ -164,3 +166,36 @@ func (tsp *TimeSeriesPoint) UnmarshalJSON(b []byte) error {
...
+		return nil
+	}
+
+	if !strings.Contains(*str, "/") && net.ParseIP(*str) != nil {
+		*str += "/32"
```
jerome-quere: do not work with ipv6

scw/custom_types.go
kindermoumoute: `{"ip_range":null}`

PR #237, opened 2019-11-13
feat(core): introduce format validation in locality parsing

QuentinBrosse:


scw/locality.go
```diff
@@ -2,10 +2,18 @@ package scw
...
 	"encoding/json"
+	"regexp"
+	"strings"
 
+	"github.com/scaleway/scaleway-sdk-go/internal/errors"
 	"github.com/scaleway/scaleway-sdk-go/logger"
 )
 
+var (
+	regionRegex = regexp.MustCompile("^[a-z]{2}-[a-z]{3}$")
```
loicbourgois: Can't we merge this with scaleway-cli.internal.validation somehow ?

We should try to have all validations in the same place.

PR #243, opened 2019-11-15
fix(config): merge selected profile on top of default profile

kindermoumoute:


scw/config.go
```diff
@@ -124,7 +124,7 @@ func (c *Config) GetProfile(profileName string) (*Profile, error) {
 		return nil, errors.New("given profile %s does not exist", profileName)
 	}
 
-	return p, nil
+	return MergeProfiles(&c.Profile, p), nil
```
jerome-quere: Add a comment 
`// Merge selected profile on top of default profil `

PR #248, opened 2019-11-20
feat(instance): add state filter on ListServers

loicbourgois:


api/instance/v1/instance_sdk.go
```diff
@@ -425,6 +425,46 @@ func (enum *ServerState) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+type ServerState string
```
jerome-quere: This is a duplicate enum. Enum already exist in Server.State in protobuf

PR #249, opened 2019-11-20
feat(rdb): wait for Instance

kindermoumoute:


api/rdb/v1/rdb_utils.go
```diff
@@ -0,0 +1,49 @@
...
+
+			if err != nil {
+				return nil, err, false
+			}
+			_, isTerminal := terminalStatus[res.Status]
+
+			return res, nil, isTerminal
+		},
+		Timeout:          req.Timeout,
+		IntervalStrategy: async.LinearIntervalStrategy(5 * time.Second),
```
loicbourgois: IMO, this should be a constant we could reuse for other `WaitFor..()` func

kindermoumoute: This constant would have to live in `scw.WaitForConstant` to avoid cross-product imports, let's keep the product logic inside the product namespace.

loicbourgois: We'll only have 1 `WaitFor..` func per product ? 

Should settings like timeouts or intervals really live inside a product ? Don't we want some consistency across all products ? 

What is the difference with rate limiting ?

PR #250, opened 2019-11-20
fix: proper ToBashArg() for IPID

loicbourgois:


strcase/bash_arg.go
```diff
@@ -5,6 +5,9 @@ import "strings"
 // ToBashArg returns the Bash public name of the given string.
 func ToBashArg(s string) string {
 	s = ToPublicGoName(s)
+	if s == "IPID" {
```
kindermoumoute: use a map `customBashName` to easily add more exceptions

PR #255, opened 2019-11-27
feat: get Region from Zone

loicbourgois:


scw/locality.go
```diff
@@ -45,6 +49,17 @@ func (zone *Zone) String() string {
 	return string(*zone)
 }
 
+// Region returns the parent Region for the Zone.
+// Manipulates the string directly to allow unlisted zones formatted as xx-yyy-z.
+func (zone *Zone) Region() (Region, error) {
```
QuentinBrosse: What do you think about

```go
func (zone *Zone) Region() (Region, error) {
    zoneStr := zone.String()
    if (validate.IsZone(zoneStr)) {
        return Region(zoneStr[:len(zoneStr)-2])
    }
    return "", fmt.Errorf("invalid zone '%v'", zoneStr)
}
```

loicbourgois: Using `IsZone()` is a good idea.

`zoneStr[:len(zoneStr)-2] ` could fail for `fr-par-12`

I prefer to work with separators than length.

QuentinBrosse: Nop, `IsZone("fr-par-12") == false`.
Hint: https://github.com/scaleway/scaleway-sdk-go/blob/master/validation/is.go#L12

loicbourgois: I'm ok with using the regex defined in validation.go to parse the Region from the zone if necessary. 

I don't think we should work with the length.

`2` looks a bit like a magic number

More generally, I think we should work with regex, then separators, then length if nothing else is available : I find it easier to read and it's more robust.

scw/locality.go
```diff
@@ -2,13 +2,17 @@ package scw
...
+	"fmt"
 	"strings"
 
 	"github.com/scaleway/scaleway-sdk-go/internal/errors"
 	"github.com/scaleway/scaleway-sdk-go/logger"
 	"github.com/scaleway/scaleway-sdk-go/validation"
 )
 
+// LocalityPartsSeparator is the separator used in Zone and Region
+const LocalityPartsSeparator = "-"
```
jerome-quere: Do not expose it publicly

scw/locality.go
```diff
@@ -45,6 +49,17 @@ func (zone *Zone) String() string {
...
 
+// Region returns the parent Region for the Zone.
+// Manipulates the string directly to allow unlisted zones formatted as xx-yyy-z.
+func (zone *Zone) Region() (Region, error) {
+	zoneStr := zone.String()
+	if !validation.IsZone(zoneStr) {
+		return "", fmt.Errorf("invalid zone '%v'", zoneStr)
+	}
+	zoneParts := strings.Split(zoneStr, LocalityPartsSeparator)
+	return Region(zoneParts[0] + LocalityPartsSeparator + zoneParts[1]), nil
```
jerome-quere: `Region(strings.Join(zoneParts[:2], "-"))`

PR #256, opened 2019-11-27
chore: rename field for update requests [APIGW-1505]

loicbourgois:


loicbourgois:
Currently, in the tests, we are using `ip.ID`for all requests.

Should we duplicate the tests, to also test with `ip.address` ? 

Or is this something that should be tested in the instance API itself ?

QuentinBrosse:
@loicbourgois In instance API itself.

PR #264, opened 2019-12-02
feat(config): add a typed error on config file not found

jerome-quere:


scw/config_test.go
```diff
@@ -265,7 +265,7 @@ func TestLoadProfileAndActiveProfile(t *testing.T) {
 		},
 		{
 			name:          "No config",
-			expectedError: "scaleway-sdk-go: cannot read config file: open {HOME}/.config/scw/config.yaml: no such file or directory",
+			expectedError: "scaleway-sdk-go: config file {HOME}/.config/scw/config.yaml could not be found",
```
kindermoumoute: nit: I would expect a message such as:
```
"scaleway-sdk-go: cannot read config file {HOME}/.config/scw/config.yaml: no such file or directory"
```
`no such file or directory` being a very common error message for shell users.

PR #273, opened 2019-12-12
feat(core): add send_usage setting in the config

jerome-quere:


scw/config.go
```diff
@@ -59,6 +59,10 @@ const configFileTemplate = `# Scaleway configuration file
 # This single default configuration is suitable for most use cases.
 {{ if .ActiveProfile }}active_profile: {{ .ActiveProfile }}{{ else }}# active_profile: myProfile{{ end }}
 
+# Help us improve your experience by automatically sending diagnostic and usage data.
+# ;-) we solemnly swear that we will only use this data to make the world a better place.
+{{ if .SendUsage }}send_usage: true{{ else }}# send_usage: false{{ end }}
```
remyleone: Should it be `{{ if .SendUsage }}send_usage: {{ .SendUsage }}{{ else }}# send_usage: false{{ end }}` ?

scw/config.go
```diff
@@ -59,6 +59,10 @@ const configFileTemplate = `# Scaleway configuration file
 # This single default configuration is suitable for most use cases.
 {{ if .ActiveProfile }}active_profile: {{ .ActiveProfile }}{{ else }}# active_profile: myProfile{{ end }}
 
+# Help us improve your experience by automatically sending diagnostic and usage data.
+# ;-) we solemnly swear that we will only use this data to make the world a better place.
```
kindermoumoute: nit:  emoji at the end
```
# We solemnly swear that we will only use this data to make the world a better place. ;-)
```

kindermoumoute: swear ==> pledge

PR #275, opened 2019-12-17
feat(strcase): add functions

remyleone:


strcase/goname_test.go
```diff
@@ -25,3 +25,23 @@ func TestLowerCaseFirstLetterOrAcronyms(t *testing.T) {
 		}
 	}
 }
+
+func TestTitleFirstWord(t *testing.T) {
+	cases := [][]string{
+		{"", ""},
+		{"t", "T"},
+		{"Test Case", "Test Case"},
+		{"test Case", "Test Case"},
```
kindermoumoute: add test `{"test case", "Test case"},`

remyleone: Fixed

PR #276, opened 2019-12-18
feat: update generated apis

kindermoumoute:
* lb: alert subscribing
* baremetal: IP failover
* k8s: v1beta4

BC: remove one of interfaces and getter


api/instance/v1/instance_sdk.go
```diff
@@ -1706,15 +1706,15 @@ func (s *API) setServer(req *setServerRequest, opts ...scw.RequestOption) (*setS
 
 type UpdateServerRequest struct {
 	Zone scw.Zone `json:"-"`
-
+	// ServerID uUID of the server
```
QuentinBrosse: nit: UUID

PR #278, opened 2019-12-20
feat(core): support more instance error types

QuentinBrosse:
Converssions:
* Error type `invalid_request_error` with `fields` field => `InvalidArgumentsError`
* Error type `invalid_request_error` without `fields` field and with `message` field => `ResponseError`
* Error type `conflict` => `ResponseError`

See `errors_test.go` for more details.

### DIFF
--- a/scw/errors.go
+++ b/scw/errors.go
@@ -160,7 +160,7 @@ func unmarshalNonStandardError(errorType string, body []byte) error {
 			return invalidArgumentsError
 		}
 
-		// At this point, the `invalid_request_error` is not an InvalidArgumentsError and
+		// At this point, the invalid_request_error is not an InvalidArgumentsError and
 		// the default marshalling will be used.
 		return nil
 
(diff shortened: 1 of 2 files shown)


scw/errors.go
```diff
@@ -185,34 +195,62 @@ func (e *InvalidArgumentsError) GetRawBody() json.RawMessage {
 	return e.RawBody
 }
 
-// InvalidRequestError are only returned by the compute API
+// InvalidRequestError is only returned by the instance API.
 // Warning: this is not a standard error.
 type InvalidRequestError struct {
+	Message string `json:"message"`
```
kindermoumoute: I don't remember seing this message field in the response, but leave it there just in case it happens. 👍

QuentinBrosse: In this kind of errors: https://github.com/scaleway/scaleway-sdk-go/pull/278/files#diff-b67d56a36244dc1228b3d114d7fdb93fR77

scw/errors.go
```diff
@@ -185,34 +195,62 @@ func (e *InvalidArgumentsError) GetRawBody() json.RawMessage {
...
+// Warning: this is not a standard error.
+type ConflictError struct {
+	Message string `json:"message"`
+
+	RawBody json.RawMessage `json:"-"`
+}
+
+// ToSdkError converts it to the standard error InvalidArgumentsError
+func (e *ConflictError) ToSdkError() SdkError {
+	return &ResponseError{
```
kindermoumoute: ToSdkError should return a standard error. The whole point of `unmarshalNonStandardError` is to convert the body ==> into a non-standard error ==> into a "faked" standard error. By default every error not catched by `unmarshalNonStandardError` and `unmarshalStandardError` will already return a `ResponseError`.

PR #283, opened 2020-01-08
chore: replace AttachIp() and DetachIP() by UpdateIP()

loicbourgois:
Internal ticket: APIGW-1716

### DIFF
--- a/api/instance/v1/instance_utils.go
+++ b/api/instance/v1/instance_utils.go
@@ -28,7 +28,7 @@ func lockServer(zone scw.Zone, serverID string) *sync.Mutex {
 
 // AttachIPRequest contains the parameters to attach an IP to a server
 //
-// Deprecated: UpdateIPRequested should be used instead
+// Deprecated: UpdateIPRequest should be used instead
 type AttachIPRequest struct {
 	Zone     scw.Zone `json:"-"`
 	IP       string   `json:"-"`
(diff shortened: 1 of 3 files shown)


api/instance/v1/instance_utils.go
```diff
@@ -83,15 +95,35 @@ type UpdateIPRequest struct {
...
 		Reverse: req.Reverse,
-	})
+	}
+
+	switch {
+	case req.Server == nil:
+		// Do nothing
+
+	case *req.Server == "":
+		// Setup server field to detach old server
```
QuentinBrosse: Are we sure we want to use an empty string to detach a server? It was one of the reasons for which we have created `DetachIP`.
If we want to keep this behavior (`""` == detach) we should document it in the method godoc.

jerome-quere: We use the same pattern for reverse already.

QuentinBrosse: Reverse use a `*NullableStringValue`. We should use the same implementation for `Reverse` and `Server` and document it in the method godoc.

loicbourgois: If we use the same pattern, why do we have
```
	Reverse *NullableStringValue `json:"reverse"`
	Server  *string              `json:"server"`
```
Do we need to change Reverse ?
Do we need to change all `*NullableStringValue` (cf internal ticket APIGW-1721) ?

PR #286, opened 2020-01-13
fix(core): handle precision and clean rounding of Money

QuentinBrosse:


scw/custom_types.go
```diff
@@ -77,12 +78,36 @@ type Money struct {
 }
 
 // NewMoneyFromFloat conerts a float with currency to a Money object.
-func NewMoneyFromFloat(value float64, currency string) *Money {
-	return &Money{
+//
+// precision: number of digits after the decimal point used to parse the nanos part of the value.
```
kindermoumoute: add examples here

PR #290, opened 2020-01-14
feat(lb): add CreateIP

QuentinBrosse:


api/lb/v1/lb_sdk.go
```diff
@@ -1578,11 +1578,55 @@ func (r *ListIPsResponse) UnsafeAppend(res interface{}) (uint32, error) {
...
+	}
+	return &resp, nil
+}
+
 type GetIPRequest struct {
 	Region scw.Region `json:"-"`
-	// IPID
-	//
-	// IP address ID.
+	// IPID iP address ID
```
kindermoumoute: lowercase `i`? 🤔

QuentinBrosse: Will be fixed in another issue.

PR #295, opened 2020-01-22
feat(instance): use a zero IntervalStrategy when replaying cassettes

kindermoumoute:


internal/async/wait.go
```diff
@@ -20,6 +21,17 @@ type WaitSyncConfig struct {
 	Timeout          time.Duration
 }
 
+// ZeroIntervalStrategy defines a zero interval duration. It is used by default when replaying cassette tests.
+func ZeroIntervalStrategy() IntervalStrategy {
+	return func() <-chan time.Time {
+		c := make(chan time.Time)
```
QuentinBrosse: `return time.After(0)`?

api/instance/v1/server_utils_test.go
```diff
@@ -13,6 +14,14 @@ import (
 	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
+func init() {
+	// set interval strategy to 0 when replaying cassettes
+	_, UpdateCassette := os.LookupEnv("UPDATE")
```
loicbourgois: shouldn't we use "UPDATE_GOLDEN" ?

kindermoumoute: This is cassette, so it would be `UPDATE_CASSETTE`

PR #307, opened 2020-02-06
feat(core): override url request option

jtherin:


scw/request_option.go
```diff
@@ -24,6 +24,13 @@ func WithAllPages() RequestOption {
 	}
 }
 
+// WithAPIURLRequest overrides the client API URL to the given URL in the request.
+func WithAPIURLRequest(apiURL string) RequestOption {
```
jerome-quere: Should we rename it to `WithRequestAPIURL` ? what do you think @QuentinBrosse

jtherin: The ability to clone a client would be nice and could replace this.

QuentinBrosse: `WithRequestAPIURL` is better yep.

QuentinBrosse:
> Another approach to this problem would be somethibng like
> 
> ```
> client := scw.NewClient()
> client2 := client.Clone(scw.WithApiURL())
> ```
> @QuentinBrosse any thought ?

@jerome-quere @jtherin 

What is the exact use-case and why [client.WithAPIURL](https://pkg.go.dev/github.com/scaleway/scaleway-sdk-go@v1.0.0-beta.6.0.20200324130151-e0b5fd35d2c8/scw?tab=doc#WithAPIURL) is not suitable?

PR #312, opened 2020-02-07
feat(k8s): add WaitForPool & WaitForClusterPools helper methods

debovema:
Requirement of https://github.com/terraform-providers/terraform-provider-scaleway/pull/393

### DIFF
--- a/api/k8s/v1beta4/k8s_helpers.go
+++ b/api/k8s/v1beta4/k8s_helpers.go
@@ -10,6 +10,7 @@ import (
 
 const (
 	waitForClusterDefaultTimeout = time.Minute * 15
+	waitForPoolDefaultTimeout    = time.Minute * 15
 )
 
 // WaitForClusterRequest is used by WaitForCluster method.
@@ -54,4 +55,80 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
 		return nil, errors.Wrap(err, "waiting for cluster failed")
 	}
 	return cluster.(*Cluster), nil
+}
+
+// WaitForClusterPoolsRequest is used by WaitForClusterPools method.
+type WaitForClusterPoolsRequest struct {
+	ClusterID string
+	Region    scw.Region
+	Timeout   *time.Duration
+}
+
+// WaitForClusterPools waits for the pools of a cluster to be ready
+func (s *API) WaitForClusterPools(req *WaitForClusterPoolsRequest) error {
+	timeout := waitForPoolDefaultTimeout
+	if req.Timeout != nil {
+		timeout = *req.Timeout
+	}
+
+	pools, err := s.ListPools(&ListPoolsRequest{
+		ClusterID: req.ClusterID,
+		Region:    req.Region,
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, pool := range pools.Pools {
+		err = s.WaitForPool(&WaitForPoolRequest{
+			PoolID:  pool.ID,
+			Timeout: &timeout,
+		})
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// WaitForPoolRequest is used by WaitForPool method.
+type WaitForPoolRequest struct {
+	PoolID  string
+	Region  scw.Region
+	Timeout *time.Duration
+}
+
+// WaitForPool waits for a pool to be ready
+func (s *API) WaitForPool(req *WaitForPoolRequest) error {
+	terminalStatus := map[PoolStatus]struct{}{
+		PoolStatusReady:   {},
+		PoolStatusWarning: {},
+	}
+
+	timeout := waitForPoolDefaultTimeout
+	if req.Timeout != nil {
+		timeout = *req.Timeout
+	}
+
+	_, err := async.WaitSync(&async.WaitSyncConfig{
+		Get: func() (interface{}, bool, error) {
+			res, err := s.GetPool(&GetPoolRequest{
+				PoolID: req.PoolID,
+				Region: req.Region,
+			})
+
+			if err != nil {
+				return nil, false, err
+			}
+			_, isTerminal := terminalStatus[res.Status]
+
+			return nil, isTerminal, nil
+		},
+		Timeout:          timeout,
+		IntervalStrategy: async.LinearIntervalStrategy(5 * time.Second),
+	})
+
+	return err
 }


api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,70 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+			return nil, false, nil
+		}
+	}
+
+	return res, true, nil
+}
+
+// WaitForPoolNodesRequest is used by WaitForPoolNodes method.
+type WaitForPoolNodesRequest struct {
+	Pool     *Pool
```
golangcibot: File is not `goimports`-ed (from `goimports`)

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,70 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
 	}
 	return cluster.(*Cluster), nil
 }
+
+// ListNodesInPool list all the nodes in a pool
+//
+// This method allows to list all the existing nodes for a specific pool.
+func (s *API) ListNodesInPool(pool *Pool) (*ListNodesResponse, error) {
```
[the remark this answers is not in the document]
QuentinBrosse: I agree

debovema:
@Sh4d1 @QuentinBrosse : thanks for your comments. I will try to have a look asap.

debovema:
> WaitForCluserNodesReady

just to be sure, ```WaitForClusterNodesReady```would just call ```WaitForPoolNodesReady``` for all the pools of the cluster ?

Sh4d1:
@debovema Since the implementation of pool ready will be to check if target size == current_node_count and the status == ready, I guess you can 
- either loop through each pool and call this method
- or check that each pool is ready and current_node_counf of cluster == sum of target size for each pool

I think both implementation are quite the same in term of performance. Maybe opt for the first one since you can in fact reuse the `WaitForPoolNodesReady` :)

debovema:
> * `WaitForNodesReadyRequest` with either a `ClusterID` or a `PoolID` and a `Region`

For this one, I am not yet as familiar as you with the Scaleway API but why do we need to provide a PoolID and a Region in the request ? It means that a PoolID is not unique across region right ? But what if we provide directly a ```*k8s.Pool``` parameter in the request ?

Sh4d1:
This method won't be implemented no? 
I guess the cleaner way is to pass the clusterID to `WaitForClusterNodesReady` and a poolID to `WaitForPoolNodesReady`. (Since you only need the ID anyway I think)

api/k8s/v1beta4/k8s_helpers.go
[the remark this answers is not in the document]
debovema: Removed

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,60 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+			return nil, false, nil
+		}
+	}
+
+	return res, true, nil
+}
+
+// WaitForPoolNodesReadyRequest is used by WaitForPoolNodesReady method.
+type WaitForPoolNodesReadyRequest struct {
+	PoolID    string
```
golangcibot: File is not `goimports`-ed (from `goimports`)

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,60 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+// WaitForPoolNodesReady waits for the nodes of a pool to be ready
+func (s *API) WaitForPoolNodesReady(req *WaitForPoolNodesReadyRequest) error {
+	pool, err := s.GetPool(&GetPoolRequest{
+		PoolID: req.PoolID,
+	})
+	if err != nil {
+		return err
+	}
+
+	r := &ListNodesRequest{
```
Sh4d1: You don't need to list the nodes, you have `pool.CurrentNodeCount` :)

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,60 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
 	}
 	return cluster.(*Cluster), nil
 }
+
+func (s *API) getNodesWhenReady(req *ListNodesRequest, expectedCount uint32) (interface{}, bool, error) {
```
Sh4d1: no need to return an interface here

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,60 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+	res, err := s.ListNodes(req, scw.WithAllPages())
```
Sh4d1: As said below, no need to list the nodes :)

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,60 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+		return nil, false, err
+	}
+
+	// first check that we have the right nodes count
+	if res.TotalCount != expectedCount {
+		return nil, false, nil
+	}
+
+	// then check that each node has Ready status
+	for _, node := range res.Nodes {
```
Sh4d1: and no need to check the status of the nodes, if the pool is ready, all the nodes in the pool are ready

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,87 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+	pools, err := s.ListPools(&ListPoolsRequest{
+		ClusterID: req.ClusterID,
+		Region:    req.Region,
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, pool := range pools.Pools {
+		s.WaitForPoolNodesReady(&WaitForPoolNodesReadyRequest{
```
golangcibot: Error return value of `s.WaitForPoolNodesReady` is not checked (from `errcheck`)

api/k8s/v1beta4/k8s_helpers.go
debovema: Are you sure of this ? I did a test using only the pool Status and it seems to be "ready" right after it is created but this does not take into account the readiness of the nodes.

Sh4d1: It does but when nodes exists. Right after pool creation, nodes does not exists, that why the combo target size == current node count + status works :) 
(You can try to check the status a few seconds  after the creation of the pool).

Sh4d1: If it doesn't work (which it should 😅) I'll fix that tomorrow :)

api/k8s/v1beta4/k8s_helpers.go
debovema: sure

api/k8s/v1beta4/k8s_helpers.go
debovema: OK so I would have to add the target size as a parameter to check on this combo ?

debovema: I tried something like
```
func (s *API) WaitForPoolNodesReady(req *WaitForPoolNodesReadyRequest) error {
	poolRequest := &GetPoolRequest{
		PoolID: req.PoolID,
	}

	_, err := async.WaitSync(&async.WaitSyncConfig{
		Get: func() (interface{}, bool, error) {
			pool, err := s.GetPool(poolRequest)
			if err != nil {
				return nil, false, err
			}

			if pool.Size == pool.CurrentNodeCount && pool.Status == "ready" { // this is immediately true after creation or update
				return nil, true, nil
			} else {
				return nil, false, nil
			}
		},
		Timeout:          req.Timeout,
		IntervalStrategy: async.LinearIntervalStrategy(5 * time.Second),
	})

	return err
}
```

and it does not work. Or do I have to provide the target size as a parameter ? If yes it's not trivial when looping in WaitForClusterNodesReady so I think I will stick with "listing" the nodes to retrieve the total count and check for every status.

Sh4d1: Hmm this should work I guess, what behaviour did you get?

debovema: The condition was true immediately even though I added ```pool.Size == pool.CurrentNodeCount```

Sh4d1: Hmm I just checked, the current node count is different than the size on creation, but status is always ready 🤔 I'll take a closer look tomorrow on this problem

debovema: Let me test again on creation since I only did test on update (1 -> 2 nodes / 2 -> 1 node)

Sh4d1: But it still won't work since the status problem 😅 should be fixed tomorrow though

debovema: Confirmed that this is the same behaviour with creation.
My Terraform output (using https://github.com/terraform-providers/terraform-provider-scaleway/pull/393) looks like this:
```
nodes = [
  {
    "public_ip" = ""
    "public_ip_v6" = ""
    "status" = "creating"
  },
]
```

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,83 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
 	}
 	return cluster.(*Cluster), nil
 }
+
+func (s *API) getNodesWhenReady(req *ListNodesRequest, expectedCount uint32) (bool, error) {
```
golangcibot: func `(*API).getNodesWhenReady` is unused (from `unused`)

api/k8s/v1beta4/k8s_helpers.go
debovema: No more nodes listing after API fix

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,62 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
 	}
 	return cluster.(*Cluster), nil
 }
+
+// WaitForClusterNodesReadyRequest is used by WaitForClusterNodesReady method.
+type WaitForClusterNodesReadyRequest struct {
+	ClusterID string
+	Region    scw.Region
+	Timeout   time.Duration
```
Sh4d1: Maybe use a `*time.Duration` and set a default value?

api/k8s/v1beta4/k8s_helpers.go
debovema: No more nodes listing after API fix

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,62 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+		}
+	}
+
+	return nil
+}
+
+// WaitForPoolNodesReadyRequest is used by WaitForPoolNodesReady method.
+type WaitForPoolNodesReadyRequest struct {
+	PoolID  string
+	Timeout time.Duration
```
Sh4d1: Add the region and the same comment as above with the timeout

api/k8s/v1beta4/k8s_helpers.go
debovema: Now using the pool status

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,62 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+	}
+
+	_, err := async.WaitSync(&async.WaitSyncConfig{
+		Get: func() (interface{}, bool, error) {
+			pool, err := s.GetPool(poolRequest)
+			if err != nil {
+				return nil, false, err
+			}
+
+			return nil, pool.Size == pool.CurrentNodeCount && pool.Status == "ready", nil
```
Sh4d1: No need to check the size and use the defined variable for the `"ready"` :)

Sh4d1: `PoolStatusReady`

api/k8s/v1beta4/k8s_helpers.go
Sh4d1: And then you can do the same as here https://github.com/scaleway/scaleway-sdk-go/blob/047c88bb15c429f5061585865410530b78e13d15/api/k8s/v1beta4/k8s_sdk.go#L988

api/k8s/v1beta4/k8s_helpers.go
debovema: Updated

api/k8s/v1beta4/k8s_helpers.go
debovema: I don't get it. Why use a pointer ? The WaitForClusterRequest above in the same file does not. And there is no default value either.

api/k8s/v1beta4/k8s_helpers.go
debovema: Region added

api/k8s/v1beta4/k8s_helpers.go
Sh4d1: Hmm fair enough 🤔
There is one here though https://github.com/scaleway/scaleway-sdk-go/blob/086aa056023ea02beba9251624b26724a212bd90/api/instance/v1/server_utils.go#L106
I'll ask @jerome-quere for his input 😄

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -47,3 +47,69 @@ func (s *API) WaitForCluster(req *WaitForClusterRequest) (*Cluster, error) {
...
+// WaitForPoolNodesReadyRequest is used by WaitForPoolNodesReady method.
+type WaitForPoolNodesReadyRequest struct {
+	PoolID  string
+	Region  scw.Region
+	Timeout time.Duration
+}
+
+// WaitForPoolNodesReady waits for the nodes of a pool to be ready
+func (s *API) WaitForPoolNodesReady(req *WaitForPoolNodesReadyRequest) error {
+	if req.Region == "" {
```
Sh4d1: Okay sorry my bad, but it's already implemented in the GetPool 😅

api/k8s/v1beta4/k8s_helpers.go
debovema: OK. AFAIK the timeout will be provided here https://github.com/terraform-providers/terraform-provider-scaleway/pull/393/files#diff-a881280cfcfd6a01417309425777e2b8R80

api/k8s/v1beta4/k8s_helpers.go
debovema: I saw it was in ListPools but didn't check in GetPool. Removing

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
+const (
+	waitForNodesReadyDefaultTimeout = time.Minute * 15
+)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
)

```

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -10,6 +10,7 @@ import (
 
 const (
 	waitForClusterDefaultTimeout = time.Minute * 15
+	waitForPoolDefaultTimeout = time.Minute * 15
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	waitForPoolDefaultTimeout    = time.Minute * 15
```

Sh4d1:
LGTM

PR #313, opened 2020-02-10
feat(core): rename send_usage into send_telemetry

QuentinBrosse:


scw/config_test.go
```diff
@@ -652,7 +652,7 @@ func TestConfig_ConfigFile(t *testing.T) {
 
 # To improve this tools we rely on diagnostic and usage data.
```
kindermoumoute: `to improve the CLI`

kindermoumoute: Or add a note `This attribute is only used for the CLI`

kindermoumoute: `we rely on diagnostic and usage data.` ==> `we rely on usage telemetry.`

scw/config_test.go
```diff
@@ -652,7 +652,7 @@ func TestConfig_ConfigFile(t *testing.T) {
...
 # Sending such data is optional and can be disable at any time by setting this variable to false.
```
kindermoumoute: Sending telemetry is optional..

scw/config.go
```diff
@@ -61,7 +61,7 @@ const configFileTemplate = `# Scaleway configuration file
 
 # To improve this tools we rely on diagnostic and usage data.
 # Sending such data is optional and can be disable at any time by setting this variable to false.
```
loicbourgois: `by setting this variable to false` -> `by setting send_telemetry to false`

PR #315, opened 2020-02-12
fix(core): handle content-types in ResponseError

QuentinBrosse:


scw/errors_test.go
```diff
@@ -27,9 +27,11 @@ func TestHasResponseErrorWithoutBody(t *testing.T) {
 
 func TestNonStandardError(t *testing.T) {
 	type testCase struct {
+		res           *http.Response
```
golangcibot: field `res` is unused (from `unused`)

PR #332, opened 2020-02-19
chore: add github action for PR testing

Sh4d1:
This PR adds GitHub actions in order to test the reliability of the solution for a few months 

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,0 +1,45 @@
+name: pull-request
+
+on:
+- pull_request
+
+jobs:
+  unit-test:
+    strategy:
+      matrix:
+        go-version: [1.10.x, 1.11.x, 1.12.x, 1.13.x]
+        platform: [ubuntu-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v1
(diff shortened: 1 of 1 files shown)


loicbourgois:
Why is this necessary ? 
Isn't CircleCI enough ?

Sh4d1:
After an irl discussion between @jth and @jerome-quere , it may be interesting to test GitHub actions in parallel with circle ci

.github/workflows/pull-request.yaml
```diff
@@ -0,0 +1,45 @@
+name: pull-request
+
+on:
+- pull_request
+
+jobs:
+  unit-test:
+    strategy:
+      matrix:
+        go-version: [1.10.x, 1.11.x, 1.12.x, 1.13.x]
```
loicbourgois: doesn't an equivalent to `golang:latest` exist ?

Sh4d1: Maybe :sweat_smile: but IMHO relying on `latest` is a very bad idea since you don't know for sure which version it will be, you don't know if you have a cached version somewhere and it's not best practice :smile: the best, in my opinion, would be to also specify the patch version but I kept the latest patch in order to be coherent with Circle CI (and today latest points to 1.13, so it's already tested)

PR #335, opened 2020-02-19
chore: remove go 1.10 and 1.11 from gh action

Sh4d1:
Go 1.10 and 1.11 are not supported anymore, so removing them from github action tests

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -7,7 +7,7 @@ jobs:
   unit-test:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -7,7 +7,7 @@ jobs:
   unit-test:
     strategy:
       matrix:
-        go-version: [1.10.x, 1.11.x, 1.12.x, 1.13.x]
+        go-version: [1.12.x, 1.13.x]
```
kindermoumoute: nit: add tip

PR #338, opened 2020-02-24
chore: untitle first word of a string

loicbourgois:


strcase/goname.go
```diff
@@ -168,6 +168,24 @@ func TitleFirstWord(s string) string {
 	return string(r)
 }
 
+// UntitleFirstWord lower case the first letter of a string.
+func UntitleFirstWord(s string) string {
+	if len(s) == 0 {
```
kindermoumoute: nit: this operation is usually done this way for string `s == ""`

kindermoumoute: This if is not required

PR #354, opened 2020-03-16
fix(k8s): add global retry interval for wait func

Sh4d1:

### DIFF
--- a/api/k8s/v1beta4/k8s_helpers.go
+++ b/api/k8s/v1beta4/k8s_helpers.go
@@ -8,9 +8,14 @@ import (
 	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
+var (
+	RetryInterval = defaultRetryInterval
+)
+
 const (
 	waitForClusterDefaultTimeout = time.Minute * 15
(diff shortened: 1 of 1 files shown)


api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -8,9 +8,14 @@ import (
 	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
+var (
+	RetryInterval = defaultRetryInterval
```
loicbourgois: please add a comment on why we need this global var

api/k8s/v1beta4/k8s_helpers.go
```diff
@@ -8,9 +8,14 @@ import (
...
+)
+
 const (
 	waitForClusterDefaultTimeout = time.Minute * 15
 	waitForPoolDefaultTimeout    = time.Minute * 15
+	defaultRetryInterval         = 5 * time.Second
```
loicbourgois:
```suggestion
	defaultRetryInterval         = time.Second * 5
```

for consistency with line above

PR #361, opened 2020-03-23
feat(error): add resource locked error

kindermoumoute:


scw/errors.go
```diff
@@ -360,6 +362,22 @@ func (e *ResourceNotFoundError) GetRawBody() json.RawMessage {
...
 
+type ResourceLockedError struct {
+	Resource   string `json:"resource"`
+	ResourceID string `json:"resource_id"`
+
+	RawBody json.RawMessage `json:"-"`
+}
+
+// IsScwSdkError implements the SdkError interface
+func (e *ResourceLockedError) IsScwSdkError() {}
```
loicbourgois: I don't understand this line 

why doesn't it do anything ?

kindermoumoute: It implements an interface

PR #366, opened 2020-04-03
feat(config): add Empty() method

kindermoumoute:


scw/config.go
```diff
@@ -143,6 +143,10 @@ func (c *Config) String() string {
 	return string(configRaw)
 }
 
+func (c *Config) Empty() bool {
```
jerome-quere, line 146: `IsEmpty` Vs `Empty` @QuentinBrosse a third opinion ?

scw/config.go
```diff
@@ -103,7 +103,7 @@ profiles:
 type Config struct {
 	Profile       `yaml:",inline"`
 	ActiveProfile *string             `yaml:"active_profile,omitempty"`
-	SendTelemetry bool                `yaml:"send_telemetry,omitempty"`
+	SendTelemetry *bool               `yaml:"send_telemetry,omitempty"`
```
jerome-quere, line 106: Does this enable telemetry by default?  So you have to explicitly opt-out?  I'm not a huge fan.
Are you sure this does not break config template generation?

kindermoumoute: No, this does not enable telemetry by default. It should actually have no impact.

It's fixing an edge case where `send_telemetry: false` is the only thing written in the config and `config.Empty()` would return true.

scw/config.go
QuentinBrosse: I vote for `IsEmpty`.

"If config is empty, do something" could be translated to:
```go
if config.IsEmpty() {
    // do something
}
```

PR #367, opened 2020-04-03
feat(config): expose GetHomeDir function

kindermoumoute:


kindermoumoute:
> This fonction should not exist and be replace with `os.GetHomeDir()`

@jerome-quere it's not listed in https://golang.org/pkg/os/

PR #386, opened 2020-04-21
feat(core): add GetSecretKey and GetAccessKey

remyleone:


scw/client.go
```diff
@@ -112,6 +112,20 @@ func (c *Client) GetDefaultZone() (zone Zone, exists bool) {
 	return Zone(""), false
 }
 
+func (c *Client) GetSecretKey() (secretKey string, exists bool) {
+	if token, isToken := c.auth.(*auth.Token); isToken {
+		return token.SecretKey, true
```
Sh4d1, line 117: can be replaced with 
```go
return token, isToken 
```
no ? same for below

PR #395, opened 2020-04-27
fix(config): LoadConfigFromPath better handle cross platform error

jerome-quere:


scw/config.go
```diff
@@ -182,11 +181,13 @@ func LoadConfig() (*Config, error) {
 
 // LoadConfigFromPath read the config from the given path.
 func LoadConfigFromPath(path string) (*Config, error) {
+	_, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return nil, configFileNotFound(path)
```
Sh4d1, line 186: Return here other types of errors too

PR #428, opened 2020-06-02
feat: add configurable retryInterval and timeout

remyleone:


remyleone:
@jerome-quere I think retryInterval should also be implemented using pointers for consistency purposes.

api/instance/v1/server_utils_test.go
```diff
@@ -16,7 +16,7 @@ import (
 func init() {
```
jerome-quere, line 16: This function is not useful anymore

PR #445, opened 2020-06-16
feat(iot): add WaitForHub IoT helper

didjcodt:
Signed-off-by: Clément Decoodt <cdecoodt@online.net>

This is highly inspired from k8s product WaitForCluster, adapted for IoT Hubs

### DIFF
--- a/api/iot/v1beta1/iot_helpers.go
+++ b/api/iot/v1beta1/iot_helpers.go
@@ -1,0 +1,63 @@
+package iot
+
+import (
+	"time"
+
+	"github.com/scaleway/scaleway-sdk-go/internal/async"
+	"github.com/scaleway/scaleway-sdk-go/internal/errors"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+const (
+	waitForHubDefaultTimeout = 15 * time.Minute
+	defaultRetryInterval     = 5 * time.Second
+)
+
+// WaitForHubRequest is used by WaitForHub method.
+type WaitForHubRequest struct {
+	HubID         string
+	Region        scw.Region
+	Status        HubStatus
+	Timeout       *time.Duration
+	RetryInterval *time.Duration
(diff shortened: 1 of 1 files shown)


api/iot/v1beta1/iot_helpers.go
```diff
@@ -0,0 +1,56 @@
...
+	HubID         string
+	Region        scw.Region
+	Status        HubStatus
+	Timeout       *time.Duration
+	RetryInterval *time.Duration
+}
+
+// WaitForHub waits for the hub to be in a ready state before returning.
+func (s *API) WaitForHub(req *WaitForHubRequest) (*Hub, error) {
+	timeout := *req.Timeout
```
remyleone, line 27: `*req.Timeout` might be nil

```suggestion
	timeout := defaultTimeout
	if req.Timeout != nil {
		timeout = *req.Timeout
	}
```

didjcodt: You're completely right, this is even done on the next parameter check. Fixed, thanks!

api/iot/v1beta1/iot_helpers.go
```diff
@@ -0,0 +1,56 @@
...
+		Get: func() (interface{}, bool, error) {
+			hub, err := s.GetHub(&GetHubRequest{
+				HubID:  req.HubID,
+				Region: req.Region,
+			})
+			if err != nil {
+				return nil, false, err
+			}
+
+			return hub, hub.Status == HubStatusReady, nil
```
jerome-quere, line 46: What is hub is in error or lock or disabled, these are final status too ?  Have a look at instance WaitServer on how we manage multiple terminal status

didjcodt: I updated to add these transient statuses too. For lock, this is not returned as a state but as an error in cases of writes on a locked hub.

PR #460, opened 2020-06-25
feat(instance): add project to resource IP

QuentinBrosse:


api/instance/v1/instance_sdk.go
```diff
@@ -4125,11 +4139,18 @@ type SetIPRequest struct {
...
 
 	Tags []string `json:"tags"`
+
+	Project string `json:"project"`
 }
 
 func (s *API) setIP(req *SetIPRequest, opts ...scw.RequestOption) (*setIPResponse, error) {
 	var err error
 
+	if req.Project == "" {
```
kindermoumoute, line 4149: this is an issue, it should be a oneof

QuentinBrosse: `setIP` is not used anymore, I will fix it in another MR

PR #543, opened 2020-09-03
doc(config): add project_id

QuentinBrosse:
We will need to update all credential links, it will be done in another PR.


kindermoumoute (approved):
LGTM

PR #886, opened 2021-08-23
feat: get metadata retry

sergelogvinov:
Try 3 times to get metadata from the cloud api.

Some times network is not ready yet, and we cannot reach the server.


remyleone:
I'm not in favor of merging this kind of work in the SDK for several reasons.
- The main one is that concurrency always depends on your particular scenario. Sometimes it will make sense to do a retry, sometimes you will want to make an exponential backoff, and so on. All this is quite far from the SDK scope and depends heavily on the problem is trying to solve.
- The second one is that if we are doing this for the metadata, we would need to maintain consistency across all the different APIs and they got very different patterns of usage/retry strategy.

PR #937, opened 2021-11-04
feat(lb-pn): add wait for attachement to load balancer

Monitob:
- Enable wait a stable status when private network is attached to a load balancer

### DIFF
--- a/api/lb/v1/lb_utils.go
+++ b/api/lb/v1/lb_utils.go
@@ -91,13 +91,12 @@ func waitForLb(timeout *time.Duration, retryInterval *time.Duration, getLB func(
 // ZonedAPIWaitForLBPNRequest is used by WaitForLBPN method.
 type ZonedAPIWaitForLBPNRequest struct {
 	LBID          string
-	PNID          string
 	Zone          scw.Zone
 	Timeout       *time.Duration
 	RetryInterval *time.Duration
 }
 
-func waitForPNLb(timeout *time.Duration, retryInterval *time.Duration, getPN func() (*PrivateNetwork, error)) (*PrivateNetwork, error) {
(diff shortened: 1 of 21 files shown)


api/lb/v1/lb_utils.go
```diff
@@ -87,3 +87,70 @@ func waitForLb(timeout *time.Duration, retryInterval *time.Duration, getLB func(
...
+	if timeout != nil {
+		t = *timeout
+	}
+	r := defaultRetryInterval
+	if retryInterval != nil {
+		r = *retryInterval
+	}
+
+	terminalStatus := map[PrivateNetworkStatus]struct{}{
+		PrivateNetworkStatusReady: {},
```
kindermoumoute, line 111: what about unknown?

api/lb/v1/lb_utils.go
```diff
@@ -87,3 +87,70 @@ func waitForLb(timeout *time.Duration, retryInterval *time.Duration, getLB func(
 	}
 	return lb.(*LB), nil
 }
+
+// ZonedAPIWaitForLBPNRequest is used by WaitForLBPN method.
+type ZonedAPIWaitForLBPNRequest struct {
+	LBID          string
+	PNID          string
```
kindermoumoute, line 94: Why do you need to wait for PN one by one? Why not wait for all PNs to be up everytime?

Monitob: Today we only support one pn but I can adapt this for future upgrades

api/lb/v1/lb_utils.go
Monitob: Not a stable state

remyleone (approved):

PR #1120, opened 2022-03-08
feat(instance): add support for tags in update snapshot

remyleone:


Monitob:
Please update the docs with remove `zone` and add `tags` from supported arguments

PR #1254, opened 2022-05-24
feat(instance) add unified volume type

Monitob:


Monitob:
I've just implemented the volume type as pointer @pierre-muller

PR #1454, opened 2022-11-07
fix(redis): return error on error state

Monitob:


remyleone:
Should be generated for all states and not manually added. I think we should close this for the time being and let the generator regenerate all of this once our generators and service definitions are updated

PR #1480, opened 2022-12-13
feat(scw): try to read yaml file as yml on fail

### LINKED ISSUE #261: Support YAML extension for the default config file name
We only support `.yaml` extension in [the default config file name](https://github.com/scaleway/scaleway-sdk-go/blob/281dd2237f297205cad5e357bdcb38d1ce26459d/scw/path.go#L14). We should also support the `.yml` extension.

This issue was reported here: https://github.com/terraform-providers/terraform-provider-scaleway/issues/292#issuecomment-559561580

Codelax:
Fix #261

### DIFF
--- a/scw/config.go
+++ b/scw/config.go
@@ -2,15 +2,18 @@ package scw
 
 import (
 	"bytes"
+	goerrors "errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"text/template"
+
+	"gopkg.in/yaml.v2"
 
 	"github.com/scaleway/scaleway-sdk-go/internal/auth"
 	"github.com/scaleway/scaleway-sdk-go/internal/errors"
 	"github.com/scaleway/scaleway-sdk-go/logger"
-	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


quantumsheep (approved):

scw/config.go
```diff
@@ -185,7 +188,22 @@ func MustLoadConfig() *Config {
...
 func LoadConfig() (*Config, error) {
-	return LoadConfigFromPath(GetConfigPath())
+	configPath := GetConfigPath()
+	cfg, err := LoadConfigFromPath(configPath)
+
+	var configNotFoundError *ConfigFileNotFoundError
+	if err != nil && goerrors.As(err, &configNotFoundError) {
+		if strings.HasSuffix(configPath, ".yaml") {
+			configPath = strings.TrimSuffix(configPath, ".yaml") + ".yml"
+		}
```
Codelax, line 198: File reading will be duplicated on fail
```suggestion
	if err != nil && goerrors.As(err, &configNotFoundError) && strings.HasSuffix(configPath, ".yaml") {
		configPath = strings.TrimSuffix(configPath, ".yaml") + ".yml"
```

quantumsheep (approved):

PR #1482, opened 2022-12-14
feat(core): move request logging from client to a transport

### LINKED ISSUE #637: Refactor logging of http requests to use an http.Transport
Refactor the logging in the SDK to use Transport for the logging.

Roundtripper can be used as a chained decorator upon an HTTP request that can implement logging, authentication management, header manipulation, ...
Beware to chain them in the right order to avoid a logging transport to leak credentials.

Codelax:
Fix #637

### DIFF
--- a/internal/auth/token.go
+++ b/internal/auth/token.go
@@ -25,7 +25,7 @@ func (t *Token) Headers() http.Header {
 	return headers
 }
 
-func AnonymizeHeaders(headers http.Header) http.Header {
+func AnonymizeTokenHeaders(headers http.Header) http.Header {
 	key := headers.Get(XAuthTokenHeader)
 	if key != "" {
 		headers.Set(XAuthTokenHeader, HideSecretKey(key))
@@ -36,7 +36,7 @@ func AnonymizeHeaders(headers http.Header) http.Header {
 // AnonymizedHeaders returns an anonymized version of Headers()
 // This method could be use for logging purpose.
 func (t *Token) AnonymizedHeaders() http.Header {
-	return AnonymizeHeaders(t.Headers())
+	return AnonymizeTokenHeaders(t.Headers())
 }
 
 func HideSecretKey(k string) string {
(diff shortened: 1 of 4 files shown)


internal/auth/token.go
```diff
@@ -25,12 +25,18 @@ func (t *Token) Headers() http.Header {
 	return headers
 }
 
+func AnonymizeHeaders(headers http.Header) http.Header {
```
Monitob, line 28: Can you rename this function may anonymizeTokenHeader

internal/auth/token.go
```diff
@@ -25,12 +25,18 @@ func (t *Token) Headers() http.Header {
...
+	key := headers.Get(XAuthTokenHeader)
+	if key != "" {
+		headers.Set(XAuthTokenHeader, HideSecretKey(key))
+	}
+	return headers
+}
+
 // AnonymizedHeaders returns an anonymized version of Headers()
 // This method could be use for logging purpose.
 func (t *Token) AnonymizedHeaders() http.Header {
```
Monitob, line 38:
```suggestion
func (t *Token) AnonymizeHeaders() http.Header {
```

internal/auth/token.go
```diff
@@ -25,12 +25,18 @@ func (t *Token) Headers() http.Header {
...
 
+func AnonymizeHeaders(headers http.Header) http.Header {
+	key := headers.Get(XAuthTokenHeader)
+	if key != "" {
+		headers.Set(XAuthTokenHeader, HideSecretKey(key))
+	}
+	return headers
+}
+
 // AnonymizedHeaders returns an anonymized version of Headers()
```
Monitob, line 36:
```suggestion
// AnonymizeHeaders returns an anonymized version of Headers()
```

internal/auth/token.go
Codelax: this return the headers of the the token but anonymiz**ed**

Monitob (approved):

PR #1483, opened 2022-12-14
fix(core): logger shouldLog checking lower log level

Codelax:
Debug = 0
Warning = 2

With level as Warning, `logger.ShouldLog(Debug)` returned true, it should return false
`Debug <= Warning` is true (wrong)
`Debug >= Warning` is false (good)


logger/default_logger.go
```diff
@@ -51,7 +52,7 @@ func (g *loggerT) Errorf(format string, args ...interface{}) {
 // ShouldLog reports whether verbosity level l is at least the requested verbose level.
 func ShouldLog(level LogLevel) bool { return logger.ShouldLog(level) }
 func (g *loggerT) ShouldLog(level LogLevel) bool {
-	return level <= g.v
+	return level >= g.v
```
quantumsheep, line 55: This was fine, why the change?

Codelax: This wasn't fine

quantumsheep (approved):

PR #1504, opened 2023-01-18
fix(marketplace/v2): format label with underscore instead of dash

Codelax:


Codelax:
Should not be done in sdk to avoid different behavior between different sdk.

PR #1535, opened 2023-02-15
feat(cockpit): implement `WaitForCockpit`

quantumsheep:


api/cockpit/v1beta1/cockpit_utils.go
```diff
@@ -0,0 +1,63 @@
...
+
+			_, isTerminal := terminalStatus[namespace.Status]
+			return namespace, isTerminal, nil
+		},
+		Timeout:          timeout,
+		IntervalStrategy: async.LinearIntervalStrategy(retryInterval),
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "waiting for DNS failed")
```
remyleone, line 59:
```suggestion
		return nil, errors.Wrap(err, "waiting for Cockpit failed")
```

api/cockpit/v1beta1/cockpit_utils.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	retryInterval := defaultRetryInterval
+	if req.RetryInterval != nil {
+		retryInterval = *req.RetryInterval
+	}
+
+	terminalStatus := map[CockpitStatus]struct{}{
+		CockpitStatusReady: {},
+		CockpitStatusError: {},
+	}
```
Vinzius, line 39: I suggest we eventually harmonize with the sdk-js/python logic, which is "wait until we see transient statutes". I think it would lead to less infinite loop in case the code or the API response is not up to date.

But probably for when we'll generate these helpers :) 

Besides that, lgtm.

Vinzius (approved):

Vinzius (approved):

PR #1794, opened 2023-08-01
feat(scw): allow usage of http proxy

cryptobioz:
This PR aims to allow usage of an HTTP proxy to communicate with the Scaleway API.

In order to keep this feature simple and standard, I suggest to use the function [ProxyFromEnvironment](https://pkg.go.dev/net/http#ProxyFromEnvironment) from the package `net/http`.

### DIFF
--- a/scw/client.go
+++ b/scw/client.go
@@ -548,6 +548,7 @@ func newHTTPClient() *http.Client {
 	return &http.Client{
 		Timeout: 30 * time.Second,
 		Transport: &http.Transport{
+			Proxy:                 http.ProxyFromEnvironment,
 			DialContext:           (&net.Dialer{Timeout: 5 * time.Second}).DialContext,
 			TLSHandshakeTimeout:   5 * time.Second,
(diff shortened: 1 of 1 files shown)


remyleone (approved):

cryptobioz:
Thanks @remyleone !

Codelax:
Could it be enough to use the custom http client option ?
```go
httpClient := &http.Client{
		Timeout: 30 * time.Second,
		Transport: &http.Transport{
			Proxy:                 http.ProxyFromEnvironment,
			DialContext:           (&net.Dialer{Timeout: 5 * time.Second}).DialContext,
			TLSHandshakeTimeout:   5 * time.Second,
			ResponseHeaderTimeout: 30 * time.Second,
			MaxIdleConnsPerHost:   20,
		},
	}
	
	client, err := scw.NewClient(scw.WithHTTPClient(httpClient))
	if err != nil {
		log.Fatalln(err)
	}
```

I would strongly favor this usage rather that adding `http.ProxyFromEnvironment` per default

PR #1916, opened 2023-11-09
feat(instance): add ipv6 metadata

agirot:
Add ipv6 api metadata to improve ip mobility support. 

Basic implementation to avoid breaking change of instance.NewMetadataAPI() signature.

### DIFF
--- a/api/instance/v1/instance_metadata_sdk.go
+++ b/api/instance/v1/instance_metadata_sdk.go
@@ -14,12 +14,15 @@ import (
 )
 
 var (
-	metadataURL           = "http://169.254.42.42"
 	metadataRetryBindPort = 200
 )
 
+const metadataAPIv4 = "http://169.254.42.42"
+const metadataAPIv6 = "http://[fd00:42::42]"
+
 // MetadataAPI metadata API
 type MetadataAPI struct {
+	MetadataURL *string
 }
 
 // NewMetadataAPI returns a MetadataAPI object from a Scaleway client.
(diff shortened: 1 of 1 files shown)


api/instance/v1/instance_metadata_sdk.go
```diff
@@ -14,22 +14,41 @@ import (
...
 
+func (meta *MetadataAPI) getMetadataUrl() string {
+	if meta.MetadataURL != nil {
+		return *meta.MetadataURL
+	}
+
+	for _, url := range []string{metadataAPIv4, metadataAPIv6} {
+		http.DefaultClient.Timeout = 3 * time.Second
+		resp, err := http.Get(url)
+		if err == nil && resp.StatusCode == 200 {
```
remyleone, line 41: What happens if err is not nil?

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -14,22 +14,41 @@ import (
...
 	return &MetadataAPI{}
 }
 
+func (meta *MetadataAPI) getMetadataUrl() string {
+	if meta.MetadataURL != nil {
+		return *meta.MetadataURL
+	}
+
+	for _, url := range []string{metadataAPIv4, metadataAPIv6} {
+		http.DefaultClient.Timeout = 3 * time.Second
```
remyleone, line 39: Could it be extracted as a variable?

remyleone (approved):

remyleone (approved):

api/instance/v1/instance_metadata_sdk.go
agirot: The next one will be use. If no IP responde same behavior like today and default v4 is return.

PR #2067, opened 2024-05-01
migrate to yaml v3

CoreFloDev:
Hi,

I am already using yaml v3 on my project, I don't see the point to have both v3 and v2 so I removed the dependence to v2 on that project. Maybe you would want to regenerate your record to have more realistic cassette, I have put 0.1ms to have the test run fast.


CoreFloDev:
Do you need the exact indentation for that test?

CoreFloDev:
Let's Get This Merged

PR #2126, opened 2024-07-02
chore: fix goreleaser format

remyleone:


.github/workflows/release.yaml
```diff
@@ -28,6 +28,6 @@ jobs:
         with:
```
Laure-di, line 28: uses: goreleaser/goreleaser-action@v6
Should not be v6?

Laure-di (approved):

PR #2227, opened 2024-09-27
fix: remove label instance image resolution

agirot:
We don't need anymore the replacement of image label to marketplace UUID.

- Instance can now take a image label and translate it into a compatible UUID
- The actual resolution don't return a compatible image with SBS and return the following error: `scaleway-sdk-go: http error 400 Bad Request: cannot create a volume of type sbs_volume from a base of type unified` on instance creation


Codelax:
Tests must be recorded as this currently run on saved requests.
I tried to run tests locally, removing local label resolution is heavily breaking.
With the following code:
```go
serverRes, err := api.CreateServer(&instance.CreateServerRequest{
		CommercialType: "DEV1-M",
		Image:          "ubuntu_jammy",
	})
	if err != nil {
		log.Fatalln(err)
	}
```
Despite creating a server with a local volume based commercial type, the created server will have a SBS volume per default as the image selected by the API is on SBS.
If next requests use instance API to interact with volumes, it will fail.

aslafy-z:
Relates to #2603

Mia-Cross:
Issue was closed by #2626

PR #2317, opened 2024-11-20
feat(rdb): fetch latest engine version from API

jremy42:


api/rdb/v1/rdb_utils.go
```diff
@@ -199,3 +199,20 @@ func (s *API) WaitForReadReplica(req *WaitForReadReplicaRequest, opts ...scw.Req
...
+	latestEngineVersion := ""
+	for _, engine := range engines.Engines {
+		if engine.Name == engineName {
+			if len(engine.Versions) > 0 {
+				latestEngineVersion = engine.Versions[0].Name
+				break
+			}
+		}
+	}
+	return latestEngineVersion, nil
```
Codelax, line 217: This should return an error if no engine is found with given name

jremy42: done

api/rdb/v1/rdb_utils.go
```diff
@@ -199,3 +200,23 @@ func (s *API) WaitForReadReplica(req *WaitForReadReplicaRequest, opts ...scw.Req
...
+func (s *API) FetchLatestEngineVersion(engineName string) (string, error) {
+	engines, err := s.ListDatabaseEngines(&ListDatabaseEnginesRequest{})
+	if err != nil {
+		return "", err
+	}
+	latestEngineVersion := ""
+	for _, engine := range engines.Engines {
+		if engine.Name == engineName {
+			if len(engine.Versions) > 0 {
+				latestEngineVersion = engine.Versions[0].Name
```
Codelax, line 213: Could you return the engine version (`engine.Versions[0]`) instead of just returning the string name of the version ?

remyleone (approved):

PR #2568, opened 2025-05-21
fix: Allow for zones as prefix in UUIDs

baalimago:
Hello!

I keep getting this warning:
```
╷
│ Warning: Using a datasource for better consistency and reliability is recommended instead of directly using offer names.
│
| ...
│
│ The offer name should be retrieved using a datasource to ensure reliable and consistent configuration.
│
│ Example:
│ data "scaleway_baremetal_offer" "my_offer_monthly" {
│   zone                = "fr-par-2"
│   name                = "EM-B112X-SSD"
│   subscription_period = "monthly"
│ }
│
│ Then, you can reference the datasource's attributes in your resources.
│
│ (and one more similar warning elsewhere)
╵
````

The reason why is since the id used returned by `scaleway_baremetal_offer` is an UUID prefix with the zone, which doesn't match the UUIDs. So this PR ensures that the IsUUID also matches on UUIDs with zone prefix.

If/once this is merged I'll submit a new PR on  updating the SDK in the provider.

### DIFF
--- a/validation/is.go
+++ b/validation/is.go
@@ -7,7 +7,7 @@ import (
 )
 
 var (
-	isUUIDRegexp  = regexp.MustCompile("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")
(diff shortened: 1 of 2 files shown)


remyleone:
I think this is much more a terraform issue than an SDK issue. The UUID regexp is fine as it is. I would suggest opening an issue on the terraform repo with a snippet of your code so that we can update the documentation to make it clearer and see where your issue is.

Laure-di:
Hello,

Thanks a lot for reporting the issue — it definitely helps us improve the provider! 🙏
I’ve just created a GitHub issue so you can follow the progress: https://github.com/scaleway/scaleway-sdk-go/issues/2569

Out of curiosity, does this warning block the creation of your resource?

Let me know if you encounter anything else!

PR #2626, opened 2025-07-02
fix(instance): resolve image label according to boot volume type

### LINKED ISSUE #2603: Wrong marketplace image selection
Using instance.CreateServer with a marketplace image like `docker`, I get 

Request:

```json
{
    "name": "instance-8c95edb7",
    "dynamic_ip_required": false,
    "commercial_type": "GP1-XS",
    "image": "65763732-a8a0-4d0a-8490-4c4e3d4682e4",
    "volumes": {
        "0": {
            "size": 15000000000,
            "volume_type": "sbs_volume"
        }
    },
    "project": "[redacted]",
    "tags": []
}
```
Response
```json
{
    "message": "Cannot create a volume of type sbs_volume from a base of type l_ssd",
    "type": "precondition_failed"
}
```

However, it works in the console:

Request
```
{
    "name": "instance-8c95edb7",
    "dynamic_ip_required": false,
    "commercial_type": "GP1-XS",
    "image": "docker",
    "volumes": {
        "0": {
            "size": 15000000000,
            "volume_type": "sbs_volume"
        }
    },
    "project": "[redacted]",
    "tags": []
}
```

Response: Success

It looks like this resolution is the culprit: https://github.com/scaleway/scaleway-sdk-go/blob/master/api/instance/v1/server_utils.go#L25-L36

This is not done by the UI, I guess the selection does not account for the volume type.

I guess the fix would be one of these:
- Let the backend do the resolution
- Update the image resolution code to select the right image

Mia-Cross:
This PR fixes the image label resolution. By first determining the type of volume on which the instance will boot, we ensure that the image fetched from the marketplace is compatible with the volume.

Closes #2603

### DIFF
--- a/api/instance/v1/server_utils.go
+++ b/api/instance/v1/server_utils.go
@@ -24,11 +24,18 @@ func (s *API) CreateServer(req *CreateServerRequest, opts ...scw.RequestOption) 
 	// If image is not a UUID we try to fetch it from marketplace.
 	if req.Image != nil && !validation.IsUUID(*req.Image) {
 		apiMarketplace := marketplace.NewAPI(s.client)
-		image, err := apiMarketplace.GetLocalImageByLabel(&marketplace.GetLocalImageByLabelRequest{
+
+		getLocalImageByLabelRequest := &marketplace.GetLocalImageByLabelRequest{
 			ImageLabel:     *req.Image,
 			Zone:           req.Zone,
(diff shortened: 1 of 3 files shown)


aslafy-z:
Thank you @Mia-Cross 
I was in fact asking for the go SDK. 
In the meantime, @remyleone made a new beta release.
Thank you for your work on this topic.

PR #3020, opened 2026-04-03
fix(locality): add support for zone `multi-az` value

estellesoulard:
`scw environmental-footprint data get` CLI call breaks in the sdk go unmarshalling, because the API sets the zone as `multi-az` for regional products, causing the following error : `scaleway-sdk-go: could not parse application/json response body: scaleway-sdk-go: bad zone format, available zones are: fr-par-1, fr-par-2, fr-par-3, nl-ams-1, nl-ams-2, nl-ams-3, pl-waw-1, pl-waw-2, pl-waw-3, it-mil-1`.

### DIFF
--- a/scw/locality.go
+++ b/scw/locality.go
@@ -37,6 +37,8 @@ const (
 	ZonePlWaw3 = Zone("pl-waw-3")
 	// ZoneItMil1 represents the it-mil-1 zone
 	ZoneItMil1 = Zone("it-mil-1")
+	// ZoneMultiAZ represents multi-az (multiple zones)
+	ZoneMultiAZ = Zone("multi-az")
 )
 
 // AllZones is an array that list all zones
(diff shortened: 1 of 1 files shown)


scw/locality.go
```diff
@@ -138,6 +140,9 @@ func ParseZone(zone string) (Zone, error) {
 		// would be triggered by API market place
 		// logger.Warningf("ams1 is a deprecated name for zone, use nl-ams-1 instead")
 		return ZoneNlAms1, nil
+	case ZoneMultiAZ.String():
+		// would be triggered by API environmental footprint
+		return ZoneMultiAZ, nil
```
estellesoulard, line 145: Handled as a specific `case` because `multi-az` appears specific to EFP API return and does not respect usual zone formatting validation. I prefer leaving it as an edge case rather than including it in our regular zones handling.

estellesoulard:
Switching to draft while discussions about the initial presence of `multi-az` value is discussed.

PR #3035, opened 2026-04-16
fix(internal/aync): fix deadlock and keep the goroutine

pierrre:
- Add a capacity to the value and error channels, so the goroutine never blocks
- Close the timeoutDone channel (it's more idiomatic)

Unfortunately we can't use the synctest package like https://github.com/scaleway/scaleway-sdk-go/pull/3034 because the goroutine is leaking.


pierrre:
Closing it in favor of https://github.com/scaleway/scaleway-sdk-go/pull/3034

PR #3100, opened 2026-06-08
feat(object): add support for s3_endpoint in profile

pypaut:
Add an `s3_endpoint` optional field in the configuration file.
Tested manually with the CLI, see https://github.com/scaleway/scaleway-cli/pull/5636.


scw/env.go
```diff
@@ -16,6 +16,7 @@ const (
 	ScwSecretKeyEnv             = "SCW_SECRET_KEY" // #nosec G101
 	ScwActiveProfileEnv         = "SCW_PROFILE"
 	ScwAPIURLEnv                = "SCW_API_URL"
+	ScwS3EndpointEnv            = "SCW_S3_ENDPOINT"
```
remyleone, line 19: How does this interact with AWS environment variables ? Would it be the same for other emulation we provide such as mnq ? 
Could you test that https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html#envvars-list-AWS_ENDPOINT_URL_SERVICE those environment variables are taken into account when used with either the terraform provider or the CLI

pypaut: `SCW_S3_ENDPOINT` is our own environment variable key. While `AWS_ENDPOINT_URL_S3` is supported, `SCW_S3_ENDPOINT` will take precedence.

For reference:

- [GetS3EndpointFromAWSConf](https://github.com/scaleway/scaleway-sdk-go/pull/3100/changes#diff-07c3a30843a3f77d46d3facec9123b804b84b5bf4d95eb84e723bb2a1b10cfffR129) (SDK)
- [getAPIEndpoint](https://github.com/scaleway/scaleway-cli/pull/5636/changes#diff-e4d629d1e15de8cf8c70ef70159ad09e345818985d1d6bea7c7796716a8820e5R444) (CLI)

The SDK exposes all the possible configuration variables, and the SDK users (CLI, Terraform provider) handle the priority logic.

estellesoulard:
Should there be some validation of the inputted s3_enpoint if it's present, similarly to apiURL in [client_option.go](https://github.com/scaleway/scaleway-sdk-go/blob/1c7c013cb7ddd8a8bf3e7fe7b9b7b3074b63eb30/scw/client_option.go#L266) ?

estellesoulard (dismissed):

scw/env.go
```diff
@@ -112,6 +118,20 @@ func LoadEnvProfile() *Profile {
...
 }
 
+// GetS3EndpointFromAWSConf retrieves the set value of AWS_ENDPOINT_URL_S3
+// or, if not set, AWS_ENDPOINT_URL.
+// This function can be called from any client side code which intends to
+// be AWS compatible, thus check the environment variables.
+// In case AWS changes the key of these variable, this function should be the
+// single point to update.
+func GetS3EndpointFromAWSConf() string {
+	if ep := os.Getenv("AWS_ENDPOINT_URL_S3"); ep != "" {
```
remyleone: Could those const be defined in scw/env.go ?

scw/env.go
```diff
@@ -112,6 +118,20 @@ func LoadEnvProfile() *Profile {
...
+		return ep
+	}
+
+	return os.Getenv("AWS_ENDPOINT_URL")
```
remyleone: Could those const be defined in scw/env.go?

remyleone (changes requested):

remyleone (approved):

PR #3254, opened 2026-08-04
feat(client): retrieve the API Gateway metadata

ds-281:
This commit adds support for retrieving the API Gateway's metadata available at (e.g.) https://api.scaleway.com/metadata.

The metadata is retrieved once, only as needed. It is then cached and the values are reused.

This paves the way for removing the hardcoded value set before the various setSRN(platform) calls.


remyleone (approved):

dmzski (approved):
LGTM

PR #3295, opened 2026-08-25
feat(instance): add new methods to pass custom ctx

### LINKED ISSUE #3290: [api/instance/v1] MetadataAPI.GetMetadata() hardcodes context.Background()
## Issue

[MetadataAPI.GetMetadata()](https://pkg.go.dev/github.com/scaleway/scaleway-sdk-go@v1.0.0-beta.37/api/instance/v1#MetadataAPI) (and its siblings ListUserData, GetUserData, SetUserData, DeleteUserData) in [api/instance/v1/instance_metadata_sdk.go](https://github.com/scaleway/scaleway-sdk-go/blob/master/api/instance/v1/instance_metadata_sdk.go) build every request with `http.NewRequestWithContext`, but the context is always a hardcoded to `context.Background()`:

```go
func (meta *MetadataAPI) GetMetadata() (m *Metadata, err error) {
        ctx := context.Background()
        req, err := http.NewRequestWithContext(ctx, http.MethodGet, meta.getMetadataURL()+"/conf?format=json", bytes.NewBufferString(""))
        ...
        resp, err := http.DefaultClient.Do(req)
        ...
}
```

This means a caller has no way to cancel or time-bound the call, not via a passed-in context.Context, and not via the client (http.DefaultClient has no Timeout set, and is a shared global besides).

## Triage

Discovered while implementing a Scaleway cloud resource detector in OpenTelemetry for opentelemetry-go-contrib; open-telemetry/opentelemetry-go-contrib#9489, where the detector needs to respect a caller-provided context but has no way to do so through this SDK method.

## Impact on downstream consumers
  
Any caller who wants a bounded/cancelable call to GetMetadata() (e.g. resource detectors, agents running on a Scaleway instance that need to give up cleanly if the metadata service is unreachable) is forced into workarounds:
- **Racing a goroutine against ctx.Done()**: this only bounds the caller's wait; the underlying GetMetadata() call and its socket keep running in the background until the OS/remote side eventually gives up, leaking a goroutine for that duration.
- **Building a private http.Client with a fixed Timeout**: bounds the leak's lifetime, but still isn't real cancellation and can't respect the caller's own deadline/cancellation semantics.
  
Both are containment, not a fix, and every consumer of this SDK's metadata API has to reinvent one of them independently.

## Suggested Fix

Thread a real context.Context through instead of hardcoding context.Background(), e.g. either:
* Add context-accepting variants (GetMetadataWithContext(ctx)) alongside the existing methods for backward compatibility, or
* Add a ctx context.Context parameter to the existing methods (breaking change, but consistent with idiomatic Go SDK design).

## Prior Art

context.Background() was introduced in #2191 purely to satisfy the noctx linter (which requires all HTTP requests to be built via NewRequestWithContext), it isn't real context propagation. The plumbing is already there, just not parameterized.

pypaut:
* Duplicate `Metadata`-related methods to `...WithContext` ones, which accept `context.Context` parameter

Fixes #3290


estellesoulard (dismissed):

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -54,6 +54,27 @@ func (meta *MetadataAPI) getMetadataURL() string {
 	return metadataAPIv4
 }
 
+func (meta *MetadataAPI) getMetadataURLWithContext(ctx context.Context) string {
```
remyleone, line 35: Instead of duplicating the code, it would be helpful to have getMetadataURL calling getMetadataURLWithContext with a context.Background(). In that way the code is not duplicated, the features are the same. The method getMetadataURL could even be deprecated.

remyleone (changes requested):

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -212,8 +215,15 @@ type Metadata struct {
 }
 
 // ListUserData returns the metadata available from the server
+//
+// Deprecated: use ListUserDataWithContext instead
+//
+//go:fix inline
 func (meta *MetadataAPI) ListUserData() (res *UserData, err error) {
-	return meta.ListUserDataWithContext(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
estellesoulard: nit: I'd move the several `3*time.Second` to a common variable and use that instead in the `context.WithTimeout` calls

pypaut: 🤌

estellesoulard (dismissed):

api/instance/v1/instance_metadata_sdk.go
ps-mir: This can cause panic in case of error, as response is nil for some of cases. Would recommend to call this conditionally for non-error scenario.

Its a practical use case for two reasons:-
* opentelemetry resource detection will run this(if used) on non-scaleway environments as a detection mechanism
* failures can happen because of other reasons.

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -32,18 +34,18 @@ func NewMetadataAPI() *MetadataAPI {
...
-func (meta *MetadataAPI) getMetadataURL() string {
+func (meta *MetadataAPI) getMetadataURLWithContext(ctx context.Context) string {
 	if meta.MetadataURL != nil {
 		return *meta.MetadataURL
 	}
 
-	ctx := context.Background()
 	for _, url := range []string{metadataAPIv4, metadataAPIv6} {
-		http.DefaultClient.Timeout = 3 * time.Second
 		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, bytes.NewBufferString(""))
```
ps-mir, line 89: Both addresses are non globally routable, and they should bypass proxy resolution entirely. This can be done at http.Transport
```go
&http.Client{Transport: &http.Transport{Proxy: nil}}
```

ps-mir (commented):
### Feedback
Need to improve cleanup for error scenario. Also the http client needs to bypass http proxy resolution entirely, since the metadata address by convention is link-local.

A unit test covering failure/success path would've caught the panic issue. I strongly recommend adding tests for a utility like this, which can have plenty of downstream customers.

### Suggestions
There might be a bit of hassle in testing this as there is no way to inject a http client, So you need to mutate `http.DefaultClient.Transport` for duration of the test, swap in a custom http.RoundTripper whose RoundTrip returns (nil, someErr).

A better path is to refactor slightly to provide option for injecting http client.

api/instance/v1/instance_metadata_sdk.go
pypaut: We now only call `Close()` when `err == nil`, which should never be an issue according to the documentation of `*http.Client.Do`:

```
If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.
```

Also, tests were added in `instance_metadata_sdk_test.go`.

api/instance/v1/instance_metadata_sdk.go
pypaut: Refactored into `newUserDataHTTPClient()`, which sets `Proxy: nil`.

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -18,51 +18,118 @@ import (
...
+
+// GetMetadataWithContext returns the metadata available from the server
+func (meta *MetadataAPI) GetMetadataWithContext(ctx context.Context) (m *Metadata, err error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, meta.getMetadataURLWithContext(ctx)+"/conf?format=json", bytes.NewBufferString(""))
 	if err != nil {
 		return nil, err
 	}
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := meta.httpClient.Do(req)
```
ps-mir, line 142: `GetMetadataWithContext` decodes response body regardless of the response status code. A non 200 json response can get decoded successfully. In such cases its hard to differentiate between
* broken metadata service silently reported as real instance with no attributes
* legitimate, but partially populated metadata.

Also a non-200 with non json body never mentions the actual status code. The error wrapping for decode and httpClient.Do is similar, making it harder to guess, if its a transport failure or bad response issue.

For bare min, a specific error type for non-200 will do the job:

```go
var ErrUnexpectedStatus = errors.New("scaleway-sdk-go: metadata service returned an unexpected status")

if resp.StatusCode != http.StatusOK {
      return nil, fmt.Errorf("%w: %d", ErrUnexpectedStatus, resp.StatusCode)
}
```

Resulting overall structure

```go
req, err := http.NewRequestWithContext(ctx, http.MethodGet, meta.getMetadataURLWithContext(ctx)+"/conf?format=json", bytes.NewBufferString(""))
if err != nil {
        return nil, err                                              // (1) can't build the request
}

resp, err := meta.httpClient.Do(req)
if err != nil {
        return nil, errors.Wrap(err, "error getting metadataURL")    // (2) transport failure: resp is nil here
}
defer resp.Body.Close()

if resp.StatusCode != http.StatusOK {
        return nil, fmt.Errorf("%w: %d", ErrUnexpectedStatus, resp.StatusCode)   // (3) NEW: resp exists, status is bad
}

metadata := &Metadata{}
if err := json.NewDecoder(resp.Body).Decode(metadata); err != nil {
        return nil, errors.Wrap(err, "error decoding metadata")       // (4) status was 200, body still garbage
}
return metadata, nil
```

ps-mir (commented):
Need specific error handling scenario to separate non-200 cases. This also helps in separating Decode based errors.

api/instance/v1/instance_metadata_sdk.go
pypaut: Thank you!

The status code is now checked in `GetMetadataWithContext`, `ListUserDataWithContext`, `GetUserDataWithContext`, `SetUserDataWithContext` and `DeleteUserDataWithContext`.

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -135,6 +141,14 @@ func (meta *MetadataAPI) GetMetadataWithContext(ctx context.Context) (m *Metadat
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf(
+			"%w: %d",
+			errors.New("metadata service returned an unexpected status"),
```
ps-mir: The Opaque error means client needs to unwrap it for error message to identify the non 200 case. 

A specific public error type for `resp.StatusCode != http.StatusOK` case would allow clients to do typed checks rather than string comparison. Ex;
`errors.Is(err, ErrUnexpectedStatus)`

Can you return a exportable error type for this case.

api/instance/v1/instance_metadata_sdk.go
```diff
@@ -305,13 +319,20 @@ func (meta *MetadataAPI) ListUserDataWithContext(ctx context.Context) (res *User
 		}
 		defer resp.Body.Close()
 
+		if resp.StatusCode != http.StatusOK {
```
ps-mir: This conflates use of retries for two orthogonal concerns.
* Transport/network failures
* Server responds, but non-200 concerns.

I would favor dropping these retries for non-200 (all recently added) entirely, and let retries focus only on transport based failures.

However in case you decide to keep it, it needs to tighten up for retriable statuses (408, 429, 502, 503, 504). A reusable function perhaps.

Recommendation is to drop the block.

ps-mir (commented):
Last few concerns.

* Need a specific Error type for non-200 case in `GetMetadataWithContext`
* non-200 based retries can be dropped.

Code is in better shape now.

ps-mir (approved):
The failure path handling is more robust now, this is ready to ship IMO.

Thanks for making all the changes.

remyleone (approved):

PR #3410, opened 2026-09-22
fix(tests): add missing requests and response filters for vcr cassettes

elwinar:


vcr/vcr.go
```diff
@@ -60,6 +55,16 @@ func cassetteRequestFilter(i *cassette.Interaction) error {
...
 		ReplaceAllString(i.Request.Body, "pack.local/builder/11111111111111111111")
 
+	for k, vs := range i.Request.Form {
+		for i, v := range vs {
+			vs[i] = regexp.MustCompile(`(.+)organization_id=[0-9a-f-]{36}(.+)`).
+				ReplaceAllString(v, "${1}organization_id=11111111-1111-1111-1111-111111111111${2}")
+			vs[i] = regexp.MustCompile(`(.+)?SCW[0-9A-Z]{17}(.+)?`).
+				ReplaceAllString(v, "${1}SCWXXXXXXXXXXXXXXXXX${2}")
+		}
+		i.Request.Form[k] = vs
```
pypaut: According to OpenCode:

```
is a no-op since vs is a slice (reference type) already mutated in place
```

vcr/vcr.go
```diff
@@ -60,6 +55,16 @@ func cassetteRequestFilter(i *cassette.Interaction) error {
 	i.Request.Body = regexp.MustCompile(`pack\.local/builder/[0-9a-f]{20}`).
 		ReplaceAllString(i.Request.Body, "pack.local/builder/11111111111111111111")
 
+	for k, vs := range i.Request.Form {
+		for i, v := range vs {
```
pypaut: Using `i` as a variable here is a bit troubling since `i` already exists outside of this scope

pypaut (changes requested):

vcr/vcr.go
elwinar: Yes, as long as we don't add anything in the slice.

vcr/vcr.go
elwinar: Understandable.

pypaut (approved):

remyleone (approved):

